Issue #19, opened 2024-03-15
Problem overriding client cache


gamab:
I wanted to have a quicker expiration time for the permissions cache.

Currently our expected approach was to use:
```go
func WithCache(cache cache.Cache) ClientOption
```

Problem1: cache is an internal package I cannot just instantiate the library localcache for my own need.

Problem2: the [cache not found error](https://github.com/grafana/authlib/blob/13ce38f75588a27b615436daf95585708f5fec1a/internal/cache/cache.go#L18) is also hidden. I cannot reuse it when implementing my own cache. Hence the client [cache error](https://github.com/grafana/authlib/blob/main/authz/client.go#L171) is always failing resulting in systematic denial of access.

PR #20, opened 2024-03-18
Cache: Make cache public

gamab:
Fix: https://github.com/grafana/authlib/issues/19


kalleep (commented):
Btw we should probably change from go-cache to something like https://github.com/dgraph-io/ristretto. Then we can control the size of the cache

gamab:
> Well right now we are not exposing a way to configure the cache even with this pr right? If you want that ability then I think authz should expose its own cache interface and expose wrapped errors?

Yes we are:
```go
func WithCache(cache cache.Cache) ClientOption
```

But we can come back on that decision and provide:
```
func WithCacheOption(cache cache.CacheOption) ClientOption
```

kalleep:
oh yeah you are right, that is strange. We are exposing a function with an interface from an internal package..

Any way do we have a case for custom implementation of the cache? Otherwise we should not expose it and just cache options I guess and when/if we have a use case for custom cache we add support for it. No need to add stuff we don't need

Jguer:
My opinion on this is that right now it's useful for us to expose it and it's getting in the way of development. Let's optimize for now here and not for when this lib is used by a 1000 integrations.

Jguer (approved):

gamab:
Ok, let's merge this then, at least the cache would be override-able.

PR #23, opened 2024-03-27
AuthN: client for access token exchange

IevaVasiljeva:
Adding a client for fetching short lived access tokens.

Depends on https://github.com/grafana/auth/pull/583

### DIFF
--- a/authn/client.go
+++ b/authn/client.go
@@ -1,0 +1,125 @@
+package authn
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"golang.org/x/sync/singleflight"
+
+	"github.com/grafana/authlib/internal/cache"
+)
+
+// Set cache TTL one minute shorter than Token expiry
+const tokenCacheTTL = 9 * time.Minute
+const tokenExchangePath = "/v1/sign-access-token"
+
+var _ TokenExchangeClient = &TokenExchangeClientImpl{}
+
+type tokenExchangeClientImpl struct {
+	cache   cache.Cache
+	cfg     Config
+	client  HTTPRequestDoer
+	singlef singleflight.Group
+}
+
+func newClient(cfg Config) (*tokenExchangeClientImpl, error) {
+	client := &tokenExchangeClientImpl{
+		client: nil,
+		cache: cache.NewLocalCache(cache.Config{
+			Expiry:          tokenCacheTTL,
+			CleanupInterval: 1 * time.Minute,
+		}),
+		cfg:     cfg,
+		singlef: singleflight.Group{},
+	}
+
+	client.client = &http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   4 * time.Second,
+				KeepAlive: 15 * time.Second,
+			}).DialContext,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       30 * time.Second,
+		},
+		Timeout: 5 * time.Second,
+	}
+
+	return client, nil
+}
+
+func (c *tokenExchangeClientImpl) GetAccessToken(ctx context.Context, req AccessTokenRequest) (string, error) {
+	key, err := tokenExchangeCacheKey(req)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate cache key: %v", err)
+	}
+	if res, err := c.cache.Get(ctx, key); err == nil {
+		token := string(res)
+		return token, nil
+	}
+
+	resp, err, _ := c.singlef.Do(key, func() (interface{}, error) {
+		target, err := url.JoinPath(c.cfg.AuthAPIURL, tokenExchangePath)
+		if err != nil {
+			return nil, err
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, target, strings.NewReader(key))
+		if err != nil {
+			return nil, err
+		}
+		req.Header.Set("Authorization", "Bearer "+c.cfg.CAP)
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Accept", "application/json")
+
+		res, err := c.client.Do(req)
+		if err != nil {
+			return nil, err
+		}
+
+		defer res.Body.Close()
+
+		response := tokenExchangeResponse{}
+		if err := json.NewDecoder(res.Body).Decode(&response); err != nil {
+			return nil, err
+		}
+
+		if res.StatusCode != http.StatusOK || response.Status != "success" {
+			return nil, fmt.Errorf("status code: %d, error: %s", res.StatusCode, response.Error)
+		}
+
+		return response, nil
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch access token: %v", err)
+	}
+
+	response, ok := resp.(tokenExchangeResponse)
+	if !ok {
+		return "", fmt.Errorf("unexpected response type")
+	}
+
+	c.cacheValue(ctx, response.Data.Token, key)
+
+	return response.Data.Token, nil
+}
+
+// TODO check what unique info for APs can be used for caching
+// TODO not enough to just cache the request probs - need to make sure that the caller should be able to get the Token
+func tokenExchangeCacheKey(query AccessTokenRequest) (string, error) {
+	data, err := json.Marshal(query)
+	return string(data), err
+}
+
+func (c *tokenExchangeClientImpl) cacheValue(ctx context.Context, token string, key string) error {
+	return c.cache.Set(ctx, key, []byte(token), cache.DefaultExpiration)
+}
--- a/authn/models.go
+++ b/authn/models.go
@@ -1,0 +1,45 @@
+package authn
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/go-jose/go-jose/v3/jwt"
+)
+
+// HTTPRequestDoer performs HTTP requests.
+// The standard http.Client implements this interface.
+type HTTPRequestDoer interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// client performs requests to auth server
+type client interface {
+	// GetAccessToken returns a short-lived access Token for the given claims.
+	GetAccessToken(ctx context.Context, req AccessTokenRequest) (string, error)
+}
+
+type TokenExchangeClient interface {
+	// GetAccessToken returns a short-lived access Token for the given claims.
+	GetAccessToken(ctx context.Context, req AccessTokenRequest) (string, error)
+}
+
+type Config struct {
+	CAP        string `json:"cloudAccessPolicy"` // cloud access policy token used for authorising the request
+	AuthAPIURL string `json:"authAPIURL"`        // URL of the auth server
+}
+
+type AccessTokenRequest struct {
+	Claims jwt.Claims     `json:"claims"` // claims to be included in the access token
+	Extra  map[string]any `json:"extra"`
+}
+
+type Data struct {
+	Token string `json:"token"`
+}
+
+type tokenExchangeResponse struct {
+	Data   Data   `json:"data"`
+	Status string `json:"status"`
+	Error  string `json:"error"`
+}
--- a/authn/token_exchange.go
+++ b/authn/token_exchange.go
@@ -1,0 +1,24 @@
+package authn
+
+import (
+	"context"
+)
+
+var _ TokenExchangeClient = &TokenExchangeClientImpl{}
+
+type TokenExchangeClientImpl struct {
+	client client
+}
+
+func NewTokenExchangeClient(cfg Config) (*TokenExchangeClientImpl, error) {
+	client, err := newClient(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return &TokenExchangeClientImpl{client: client}, nil
+}
+
+func (s *TokenExchangeClientImpl) GetAccessToken(ctx context.Context, req AccessTokenRequest) (string, error) {
+	return s.client.GetAccessToken(ctx, req)
+}
--- a/authn/client_test.go
+++ b/authn/client_test.go
@@ -1,0 +1,60 @@
+package authn
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestClientImpl_Search(t *testing.T) {
+	tests := []struct {
+		name     string
+		response string
+		want     string
+		wantErr  bool
+	}{
+		{
+			name:     "Can parse a successful response",
+			response: `{"status":"success","data":{"token":"exchanged_token"}}`,
+			want:     "exchanged_token",
+			wantErr:  false,
+		},
+		{
+			name:     "Can parse an error response",
+			response: `{"status":"error","error":"invalid permission"}`,
+			want:     "",
+			wantErr:  true,
+		},
+	}
+	for _, tt := range tests {
+		capToken := "test_cap_token"
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			require.Equal(t, r.Header.Get("Authorization"), fmt.Sprintf("Bearer %s", capToken))
+			require.Equal(t, r.URL.Path, tokenExchangePath)
+			_, _ = w.Write([]byte(tt.response))
+
+		}))
+		defer server.Close()
+		t.Run(tt.name, func(t *testing.T) {
+			c, err := newClient(Config{
+				AuthAPIURL: server.URL,
+				CAP:        capToken,
+			})
+			require.NoError(t, err)
+
+			c.client = server.Client()
+
+			token, err := c.GetAccessToken(context.Background(), AccessTokenRequest{})
+			if tt.wantErr {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, tt.want, token)
+		})
+	}
+}


authn/client.go
```diff
@@ -0,0 +1,125 @@
...
+		client: nil,
+		cache: cache.NewLocalCache(cache.Config{
+			Expiry:          tokenCacheTTL,
+			CleanupInterval: 1 * time.Minute,
+		}),
+		cfg:     cfg,
+		singlef: singleflight.Group{},
+	}
+
+	client.client = &http.Client{
```
IevaVasiljeva, line 42: I'm reusing the same settings that are used by authZ client, as they seemed like sensible defaults.

authn/client.go
```diff
@@ -0,0 +1,125 @@
...
+			IdleConnTimeout:       30 * time.Second,
+		},
+		Timeout: 5 * time.Second,
+	}
+
+	return client, nil
+}
+
+func (c *tokenExchangeClientImpl) GetAccessToken(ctx context.Context, req AccessTokenRequest) (string, error) {
+	key, err := tokenExchangeCacheKey(req)
```
IevaVasiljeva, line 61: I'm not sure what the cache should be based on. Do we expect the users to specify anything in the request?

authn/models.go
```diff
@@ -0,0 +1,45 @@
...
+	"net/http"
+
+	"github.com/go-jose/go-jose/v3/jwt"
+)
+
+// HTTPRequestDoer performs HTTP requests.
+// The standard http.Client implements this interface.
+type HTTPRequestDoer interface {
+	Do(req *http.Request) (*http.Response, error)
+}
```
IevaVasiljeva, line 14: Copy paste from the authZ client - perhaps we should create some utility package?

authn/token_exchange.go
IevaVasiljeva, line 1: I was following the same interface pattern as we have for authZ client, but I'm not sure it makes sense here, as the only functionality that we expose to users is token exchange, and the only thing that the client does is token exchange. So perhaps we can remove one of the layers of abstraction. On the other hand, maybe it makes sense to keep them if we plan to extend the client in the future.

authn/client.go
Jguer: Normally the cache should be only based on the `exp` of the access token.

Jguer: and the audience I think

authn/models.go
```diff
@@ -0,0 +1,45 @@
...
+	// GetAccessToken returns a short-lived access Token for the given claims.
+	GetAccessToken(ctx context.Context, req AccessTokenRequest) (string, error)
+}
+
+type Config struct {
+	CAP        string `json:"cloudAccessPolicy"` // cloud access policy token used for authorising the request
+	AuthAPIURL string `json:"authAPIURL"`        // URL of the auth server
+}
+
+type AccessTokenRequest struct {
```
Jguer, line 32: I think you can add Realm and OrgID to this request, for the cases when a system token is used

On the API call they can be set as the following headers:

| X-Org-ID | string | Example: 1 | The ID of the org the request should be restricted to. Required when using a system token. |
| -- | -- | -- | -- |
| X-Realms | string | Example: [{"type":"org", "identifier": "1"}] | A JSON-encoded array of objects containing the realms the request should be restricted to. Required when using a system token. |

authn/models.go
```diff
@@ -0,0 +1,45 @@
+package authn
```
kalleep, line 1: I am not a big fan of model files in general. I think we should define interfaces / structures where they are used

authn/models.go
kalleep: I don't see the need for this interface. we can just use *http.Client directly. We don't use it for tests or expose a what to override this

authn/token_exchange.go
kalleep: No I think you are spot on. We don't need the extra level of indirection we have right now

authn/models.go
IevaVasiljeva: Yeah, I think it's more for future proofing to make tests easier to write. But I will remove it now and we can always add something like that back in when it's actually needed.

IevaVasiljeva:
Thanks for your reviews @kalleep and @Jguer! I've made a bunch of changes, and would appreciate another review.

The main changes:
* simplify the structure to remove unneeded interface layers and not have a separate models file;
* change the request struct to take in realms and org ID, and not require claims for now;
* check if the CAP that is used is system wide, and require realms and org ID in that case;
* parse the received token (for successful requests) and use token's expiry time as TTL for the cache.

authn/token_exchange.go
```diff
@@ -0,0 +1,211 @@
...
+
+const cacheBuffer = 15 * time.Second
+const tokenExchangePath = "/v1/sign-access-token"
+
+type TokenExchangeClient interface {
+	// GetAccessToken returns a short-lived access Token for the given claims.
+	GetAccessToken(ctx context.Context, req AccessTokenRequest) (string, error)
+}
+
+type Config struct {
```
kalleep, line 29:
```suggestion
type TokenEnchangeConfig struct {
```

kalleep (commented):
Looks much better, left a nit and a question

authn/token_exchange.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+type Realm struct {
+	Type       string `json:"type"`       // org or stack
+	Identifier string `json:"identifier"` // org id or stack id
+}
+
+type AccessTokenRequest struct {
+	Realms []Realm // A JSON-encoded array of objects containing the realms the request should be restricted to.
+	OrgID  int64   // ID of the org the request should be restricted to.
```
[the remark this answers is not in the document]
IevaVasiljeva: I'm not sure about it, I think it might be needed if stack is specified in the realms instead of an org. Otherwise you're right that we could extract org ID from the realms.

authn/token_exchange.go
IevaVasiljeva: Thanks, good point. I'll rename.

authn/token_exchange.go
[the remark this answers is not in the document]
kalleep: Maybe it would be good to get some eyes on this from someone that knows more about how system tokens work, like @drclau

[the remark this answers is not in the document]
cinaglia: The API does currently require `X-Org-ID` to always be set even if the realm is of type `org`. This is done as an extra layer of security such that we're able to prevent the scenario where a user picks an `org` realm it does not have permissions over.

authn/token_exchange.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+type tokenExchangeResponse struct {
+	Data   TokenExchangeData `json:"data"`
+	Status string            `json:"status"`
+	Error  string            `json:"error"`
+}
+
+func (c *tokenExchangeClientImpl) GetAccessToken(ctx context.Context, tokenReq TokenExchangeRequest) (string, error) {
+	if c.cfg.usesSystemCAP && (tokenReq.OrgID == 0 || len(tokenReq.Realms) == 0) {
```
cinaglia, line 123: I don't have a really strong preference here but FWIW we should be able to hide the concept of system tokens from the client:

- When using a system token (OBO flow) w/o providing `X-Org-ID` and `X-Realms`, the API will return a 401 error.
- When using a user token, both headers are ignored.

Perhaps an idea would be to have two entrypoints? One for the regular use-case, and another for the on-behalf-of flow, where both OrgID and Realms are required? Alternatively, we could simply keep the same signature and make `tokenReq` optional.

authn/token_exchange.go
```diff
@@ -0,0 +1,211 @@
...
+const cacheBuffer = 15 * time.Second
+const tokenExchangePath = "/v1/sign-access-token"
+
+type TokenExchangeClient interface {
+	// GetAccessToken returns a short-lived access Token for the given claims.
+	GetAccessToken(ctx context.Context, req TokenExchangeRequest) (string, error)
+}
+
+type TokenExchangeConfig struct {
+	CAP           string // cloud access policy token used for authorising the request
```
cinaglia, line 30: nit: This is actually a cloud access policy token.

authn/token_exchange.go
```diff
@@ -0,0 +1,211 @@
...
+			return nil, err
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, target, strings.NewReader("{}"))
+		if err != nil {
+			return nil, err
+		}
+		req.Header.Set("Authorization", "Bearer "+c.cfg.CAP)
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Accept", "application/json")
```
cinaglia, line 148: suggestion/optional: It might be nice to set an `User-Agent` header for all requests made from the client.

authn/token_exchange.go
```diff
@@ -0,0 +1,211 @@
...
+	Type       string `json:"type"`       // org or stack
+	Identifier string `json:"identifier"` // org id or stack id
+}
+
+type TokenExchangeRequest struct {
+	Realms []Realm // A JSON-encoded array of objects containing the realms the request should be restricted to.
+	OrgID  int64   // ID of the org the request should be restricted to.
+}
+
+type TokenExchangeData struct {
```
cinaglia, line 112: Do we need to keep this public if `(tokenExchangeClientImpl).GetAccessToken` returns a `string`?

authn/token_exchange.go
```diff
@@ -0,0 +1,211 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch access token: %v", err)
+	}
+
+	response, ok := resp.(tokenExchangeResponse)
+	if !ok {
+		return "", fmt.Errorf("unexpected response type")
+	}
+
+	c.cacheValue(ctx, response.Data.Token, key)
```
cinaglia, line 187: Did you intend to ignore the error here? I suppose there's a little bit of value in handling the error since it would enforce that the received token is a valid JWT.

cinaglia (commented):
This looks great. I just had a couple of minor comments.

authn/token_exchange.go
charandas: I support the idea of not allowing arbitrary realms for `system` CAP in general but currently those 2 headers are used to determine aud. I am wondering how would one go about setting an aud in the usual oauth sense: where it's usually a URL for the receiving service.

I am specifically thinking of multi-tenant services as destination (also the source) of one of these exchanged tokens.

kalleep: >I am specifically thinking of multi-tenant services as destination (also the source) of one of these exchanged tokens.

Going multi-tenant -> single-tenant grafana instance is fine you would generate a access token where aud would be the stack id and that can then be validated. E.g. grafana would only accept tokens that has their stack id in aud claim. I wonder how this should look like for multi-tenant -> mutlti-tenant. Having a stack id in aud could limit the request to that service to limit request to a specific stack, same if org / tenant id is used I guess. But there could be value to be able to sign a access token that is not limited to those

authn/token_exchange.go
IevaVasiljeva: Yup, you're right, this struct can be private.

authn/token_exchange.go
IevaVasiljeva: I was thinking about it, but it felt like it's not worth returning an error if we simply fail to cache the token. I've since refactored the code, so `cacheValue ` would also return an error if the token is not a valid JWT, so I think you're right that it makes sense to return an error instead of swallowing it. I'll do that for now, and we can change it later if needed.

authn/token_exchange.go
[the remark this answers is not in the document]
IevaVasiljeva: I take it back - I read through the conversation [here](https://github.com/grafana/authlib/pull/23#discussion_r1553524613) again, and tried to follow Victor's suggestion for having two different entrypoints for system token and user token. Please TAL and let me know if I've misunderstood the suggestion.

cinaglia (approved):
Looks great to me! Will defer to @kalleep as I believe he had left some comments in his last review.

authn/token_exchange.go
```diff
@@ -0,0 +1,198 @@
...
+	client := &tokenExchangeClientImpl{
+		client: nil,
+		cache: cache.NewLocalCache(cache.Config{
+			CleanupInterval: 5 * time.Minute,
+		}),
+		cfg:     cfg,
+		singlef: singleflight.Group{},
+	}
+
+	client.client = &http.Client{
```
charandas, line 53: Is it possible for you to allow us to provide the http.Client itself? Asking for Grafana where we are using a clientProvider interface from grafana-plugin-sdk which has some nice defaults as well as tracing and other config preset.

charandas: Maybe as an optional to `NewTokenExchangeClient`?

authn/token_exchange.go
```diff
@@ -0,0 +1,198 @@
...
+type tokenExchangeResponse struct {
+	Data   tokenExchangeData `json:"data"`
+	Status string            `json:"status"`
+	Error  string            `json:"error"`
+}
+
+func (c *tokenExchangeClientImpl) GetAccessTokenOnBehalfOf(ctx context.Context, tokenReq TokenExchangeRequest) (string, error) {
+	if tokenReq.OrgID == 0 || len(tokenReq.Realms) == 0 {
+		return "", fmt.Errorf("org ID and realms must be specified when fecthing access token on behalf of a user")
+	}
```
kalleep, line 101: I am not sure this is the correct name, same for the error. I don't think omitting `X-Realms` and `X-Org-Id` would work for system tokens right now when signing but it should work. I think right now `GetAccessToken` would fetch an access token without providing any headers e.i. use whatever org id and realms is configured in CAP.

For system tokens I think we need to make it a "system" access token, i.e. the token that gets singed is multi-tenant so that it is not restricted to a stack/org. This method (`GetAccessTokenOnBehalfOf`) is a way to request access token that is restricted to one or more of these and is only really usable with a system token. So it is not for request a "OnBehalfOf" token but to request a more restricted access token when you have a system token

authn/token_exchange.go
IevaVasiljeva: Yup, can do 👍

kalleep:
Thinking a bit more about this I think we should probably have two different clients. One to use when you have a system token and one to use when you have a non system token, they can share an interface but don't necessarily have to do it.

Users of this component should know if they need to use a system token or a token with non system realms. Then we would only need one method on each, `Exchange`.

For `SystemTokenExchangeClient` we would add `X-Org-Id` and `X-Realms` headers if they are provided. For `RealmTokenExchangeClient` we would never add those header to a request. This would avoid the confusing function names and not having a function that only works if you have a system token

authn/token_exchange.go
```diff
@@ -0,0 +1,216 @@
...
+
+type tokenExchangeResponse struct {
+	Data   tokenExchangeData `json:"data"`
+	Status string            `json:"status"`
+	Error  string            `json:"error"`
+}
+
+func (c *tokenExchangeClientImpl) ExchangeSystemToken(ctx context.Context, tokenReq TokenExchangeRequest) (string, error) {
+	if tokenReq.OrgID == 0 || len(tokenReq.Realms) == 0 {
+		return "", fmt.Errorf("org ID and realms must be specified when fecthing access token on behalf of a user")
```
kalleep, line 118:
```suggestion
		return "", fmt.Errorf("org ID and realms must be specified when fecthing access token")
```

authn/token_exchange.go
```diff
@@ -0,0 +1,216 @@
...
+type ClientOption func(impl *tokenExchangeClientImpl)
+
+// WithHTTPClient allows setting the HTTP client to be used by the token exchange client.
+func WithHTTPClient(client *http.Client) ClientOption {
+	return func(c *tokenExchangeClientImpl) {
+		c.client = client
+	}
+}
+
+func NewTokenExchangeClient(cfg TokenExchangeConfig, opts ...ClientOption) (*tokenExchangeClientImpl, error) {
```
kalleep, line 48: Right now users would get a private structure that implement both `RealmTokenExchangeClient` and `SystemTokenExchangeClient`. I don't think we ever need that. 

If we want to keep the implementation in the same structure we should have one constructor for each that returns the interface instead of the concrete type. So I think we need to have two different constructors so users can chose which one to use.

The other option would be to have two different structures that we expose. One for each interface and return the concrete type of those in constructors instead

kalleep (commented):
Nice, I think this is a better direction. I am curious to see that @cinaglia think of doing this. This is not really abstracting away "system" tokens but I think it makes is easier to work with these components

authn/token_exchange.go
IevaVasiljeva: Makes sense, I've split the split the constructors up.

authn/token_exchange.go
```diff
@@ -0,0 +1,224 @@
...
+
+type tokenExchangeResponse struct {
+	Data   tokenExchangeData `json:"data"`
+	Status string            `json:"status"`
+	Error  string            `json:"error"`
+}
+
+func (c *tokenExchangeClientImpl) ExchangeSystemToken(ctx context.Context, tokenReq TokenExchangeRequest) (string, error) {
+	if tokenReq.OrgID == 0 || len(tokenReq.Realms) == 0 {
+		return "", fmt.Errorf("org ID and realms must be specified when fecthing access token for a system token")
```
charandas, line 126:
```suggestion
		return "", fmt.Errorf("org ID and realms must be specified when fetching access token for a system token")
```

authn/token_exchange.go
charandas: There is definite value to signing a token not limited to a stack/org. I hope you can relax the condition on exchange endpoint so that a system CAP can exchange for a token of that kind.

Maybe this PR can also relax that requirement and simply let the backend API decide what the semantics are?

charandas: I apologize if you have already answered my request in adding `ExchangeRealmToken`. I was just confused by the error message about providing those other 2 headers in the other function.

kalleep:
@charandas @cinaglia I did significant changes to this pr in  [4979813](https://github.com/grafana/authlib/pull/23/commits/4979813b19f5d45a5584120cee9d3014dbcbf965) that greatly simplified things.

As we discussed in https://github.com/grafana/auth/pull/613 we always set `X-Org-Id=0` and `X-Realms=[{"type": "system", "identifier": "system"}]`. For none system token these will always be ignored. Instead of having different functions depending on token realm type we can now have one function for both cases. To perform an exchange request we require  that namespace is set and that we have at least one audience. 

So even if a service is using a non system token they still need to pass a namspace.

authn/token_exchange.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	key := r.hash()
+	token, ok := c.getCache(ctx, key)
+	if ok {
+		return &TokenExhangeResponse{Token: token}, nil
+	}
+
+	resp, err, _ := c.singlef.Do(key, func() (interface{}, error) {
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.cfg.TokenExchangeURL, strings.NewReader("{}"))
```
cinaglia, line 113: It seems as though you forgot to marshal `r` and provide that as the request body.

authn/token_exchange.go
```diff
@@ -0,0 +1,184 @@
...
+
+type TokenExhangeResponse struct {
+	Token string
+}
+
+func (r TokenExchangeRequest) hash() string {
+	br := strings.Builder{}
+	br.WriteString(r.Namespace)
+	br.WriteByte('-')
+	br.WriteString(strings.Join(r.Audiences, "-"))
```
cinaglia, line 82: nit/optional: There might be value in sorting the list of audiences here to make the hash more predictable.

cinaglia (approved):

charandas:
Shouldn't `namespace` be added here: https://github.com/grafana/authlib/blob/3e6125d0a52c1cf68b05fa511b8d3bf8ffbd29a0/authn/verifier_access_token.go#L11?

authn/token_exchange.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (c *TokenExchangeClient) withHeaders(r *http.Request) *http.Request {
+	r.Header.Set("Authorization", "Bearer "+c.cfg.Token)
+	r.Header.Set("Content-Type", "application/json")
+	r.Header.Set("Accept", "application/json")
+	r.Header.Set("User-Agent", "authlib-client")
+
+	// Always propagate system token headers.
+	// These will be ignored for non system tokens.
+	r.Header.Set("X-Org-ID", "0")
```
charandas, line 168: I would expect no problems since these are always being set.

authn/token_exchange.go
```diff
@@ -0,0 +1,194 @@
...
+	r.Header.Set("X-Realms", `{"type": "system", "identifier": "system"}`)
```
cinaglia, line 169:
```suggestion
	r.Header.Set("X-Realms", `[{"type": "system", "identifier": "system"}]`)
```

PR #25, opened 2024-04-02
Authn: Add common wrappers for identity and access tokens

kalleep:
This add two wrappers around verifier with common use-cases we have.

One wrapper for extrating and validating grafana id tokens and one for access tokens. I also break the api on verify to include the expected typ of the token. We use `at+jwt` for access tokens and `jwt` for id tokens.

### DIFF
--- a/authn/extractor_access.go
+++ b/authn/extractor_access.go
@@ -1,0 +1,67 @@
+package authn
+
+import (
+	"context"
+	"strings"
+
+	"github.com/go-jose/go-jose/v3/jwt"
+	"github.com/grafana/authlib/cache"
+)
+
+const (
+	tenantPrefixOrg   = "org:"
+	tenantPrefixStack = "stack:"
+)
+
+type accessClaims struct {
+	jwt.Claims
+	Scopes      []string `json:"scopes"`
+	Permissions []string `json:"permissions"`
+}
+
+type AccessToken struct {
+	// TenantID is an prefixed tenant identitfier either `org:` or `stack:`.
+	TenantID string
+	// Access policy scopes
+	Scopes []string `json:"scopes"`
+	// Grafana roles
+	Permissions []string `json:"permissions"`
+}
+
+func NewAccessExtractor(cfg IDVerifierConfig) *AccessExtractor {
+	return &AccessExtractor{
+		v: NewVerifier[accessClaims](cfg),
+	}
+}
+
+func NewAccessExtractorWithCache(cfg IDVerifierConfig, cache cache.Cache) *AccessExtractor {
+	return &AccessExtractor{
+		v: newVerifierWithKeyService[accessClaims](cfg, newKeyServiceWithCache(cfg.SigningKeysURL, cache)),
+	}
+}
+
+type AccessExtractor struct {
+	v Verifier[accessClaims]
+}
+
+func (e *AccessExtractor) FromToken(ctx context.Context, token string) (*AccessToken, error) {
+	claims, err := e.v.Verify(ctx, token, TypeAccessToken)
+	if err != nil {
+		return nil, err
+	}
+
+	access := &AccessToken{
+		Scopes:      claims.Rest.Scopes,
+		Permissions: claims.Rest.Permissions,
+	}
+
+	for _, aud := range claims.Audience {
+		if strings.HasPrefix(aud, tenantPrefixOrg) || strings.HasPrefix(aud, tenantPrefixStack) {
+			access.TenantID = aud
+			break
+		}
+
+	}
+
+	return access, nil
+}
--- a/authn/verifier.go
+++ b/authn/verifier.go
@@ -9,13 +9,16 @@ import (
 	"github.com/go-jose/go-jose/v3/jwt"
 )
 
+const (
+	TypeIDToken     = "jwt"
+	TypeAccessToken = "at+jwt"
+)
+
 type Verifier[T any] interface {
 	// Verify will parse and verify provided token using public key from `IDVerifierConfig.SigningKeysURL`.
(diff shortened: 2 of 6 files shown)


authn/extractor_access.go
```diff
@@ -0,0 +1,67 @@
...
+
+type accessClaims struct {
+	jwt.Claims
+	Scopes      []string `json:"scopes"`
+	Permissions []string `json:"permissions"`
+}
+
+type AccessToken struct {
+	// TenantID is an prefixed tenant identitfier either `org:` or `stack:`.
+	TenantID string
```
kalleep, line 24: Suggestions for a better name here? Same for Identity struct I guess

authn/extractor_access.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	access := &AccessToken{
+		Scopes:      claims.Rest.Scopes,
+		Permissions: claims.Rest.Permissions,
+	}
+
+	for _, aud := range claims.Audience {
+		if strings.HasPrefix(aud, tenantPrefixOrg) || strings.HasPrefix(aud, tenantPrefixStack) {
+			access.TenantID = aud
```
cinaglia, line 60: Access policies may have (and often do) have more than one realm. For instance, an access policy that can be used for managing resources belonging specifically to stack A and B. By breaking early after finding the first tenant, we cannot guarantee the tenant ID will be correct.

See https://github.com/grafana/auth/blob/ae191ac194542bbfcbc4892e044ad7e5ea22b754/pkg/tokensigner/token_signer.go#L123-L129 and https://github.com/grafana/auth/blob/ae191ac194542bbfcbc4892e044ad7e5ea22b754/pkg/tokensigner/token_signer.go#L157-L161. The signed tokens will have all realms.

authn/extractor_access.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	tenantPrefixOrg   = "org:"
+	tenantPrefixStack = "stack:"
+)
+
+type accessClaims struct {
+	jwt.Claims
+	Scopes      []string `json:"scopes"`
+	Permissions []string `json:"permissions"`
```
cinaglia, line 19: Should `delegatedPermissions` be included here?

authn/verifier.go
```diff
@@ -19,20 +27,30 @@ type Claims[T any] struct {
...
 
 type VerifierBase[T any] struct {
 	cfg  IDVerifierConfig
 	keys *keyService
 }
 
-func (v *VerifierBase[T]) Verify(ctx context.Context, token string) (*Claims[T], error) {
+// Verify will parse and verify provided token using public key from `SigningKeysURL`.
+// If `AllowedAutiences` was configured those will be validated as well.
+func (v *VerifierBase[T]) Verify(ctx context.Context, token, typ string) (*Claims[T], error) {
```
cinaglia, line 44: Would things be cleaner if we were to move the type to `IDVerifierConfig`? I could be wrong but it feels as though the scenario of a verifier having to be able to validate more than one type is unlikely. In which case, I don't think developer experience would be affected if users had to create multiple verifier instances per type.

kalleep: Yeah I think that would be better. It was my first thought but I don't remember why I did not end up doing so

authn/extractor_access.go
kalleep: Yep you are right, I took these from https://github.com/grafana/grafana/pull/83814/files#diff-116e2e2a9f1c93fe98ca1f357ef33dba2b473838c913e1905140750ad56289abR61-R69 and missed DelegatedPermissions

authn/verifier.go
cinaglia: Yeah. It seems like it. If push comes to shove, we could write a helper function that facilitates the tenant validation (e.g. to check whether the tenant list includes the relevant tenant).

kalleep:
@cinaglia  I refactor it quite a lot, now they are truly just convenient wrappers for access token and id token. They Instead return expected claims, so not doing anything with the audience.

I also moved the type to creation. I was thinking of adding it to config first but not sure it is needed from there. They annoying thing will be to sync these claims when we do changes. But the alternative would be that everyone that uses Verifier directly would have to do it themself. If these wrappers are used it would be done by upgrading the package

authn/verifier_id_token.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+func NewIDTokenVerifierWithCache(cfg VerifierConfig, cache cache.Cache) *IDTokenVerifier {
+	return &IDTokenVerifier{
+		v: newVerifierWithKeyService[IDTokenClaims](cfg, TokenTypeID, newKeyServiceWithCache(cfg.SigningKeysURL, cache)),
+	}
+}
+
+// IDTokenVerifier is a convient wrapper around `Verifier`
```
cinaglia, line 26:
```suggestion
// IDTokenVerifier is a convenient wrapper around `Verifier`
```

authn/verifier_access_token.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+}
+
+func NewAccessTokenVerifierWithCache(cfg VerifierConfig, cache cache.Cache) *AccessTokenVerifier {
+	return &AccessTokenVerifier{
+		v: newVerifierWithKeyService[AccessTokenClaims](cfg, TokenTypeAccess, newKeyServiceWithCache(cfg.SigningKeysURL, cache)),
+	}
+}
+
+// AccessTokenVerifier is a convient wrapper around `Verifier`
```
cinaglia, line 30:
```suggestion
// AccessTokenVerifier is a convenient wrapper around `Verifier`
```

cinaglia (approved):
Looks great!

PR #27, opened 2024-05-02
POC changes to allow provisioning for ST bound tokens using authlib

charandas:
While the API allows for this, the library is blocking provisioning the older (still valid) kind of ST tokens. Here, I have tried to work this out using the namespace value already present.

It goes on to show though that another parameter for orgID would need to be added when namespace isn't `*`. That, unless we relax the requirement for orgId at the API.


charandas:
Based on my new understanding, that I can sign with a static `aud` when tokens are outbound for Grafana, I think I don't need this. Closing.

PR #36, opened 2024-05-31
Authn: Allow configuring the key retriever http client

gamab:
In the scope of the `UniStore` work, given `Grafana` is relying on a self signed https certificate for local testing, I'd like to be able to replace the client with one that has `tls.Config{InsecureSkipVerify: true}`

### DIFF
--- a/authn/jwks.go
+++ b/authn/jwks.go
@@ -42,7 +42,7 @@ func NewKeyRetriever(cfg KeyRetrieverConfig, opt ...DefaultKeyRetrieverOption) *
 	return s
 }
 
-// WithHTTPClientKeyRetrieverOpt allows setting the HTTP client to be used by the token exchange client.
+// WithHTTPClientKeyRetrieverOpt allows setting the HTTP client to be used by the key retriever.
 func WithHTTPClientKeyRetrieverOpt(client *http.Client) DefaultKeyRetrieverOption {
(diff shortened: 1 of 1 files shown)


authn/jwks.go
```diff
@@ -18,26 +18,42 @@ type KeyRetriever interface {
...
+	}
+
+	for _, o := range opt {
+		o(s)
+	}
+	return s
+}
+
+// WithHTTPClientKeyRetrieverOpt allows setting the HTTP client to be used by the token exchange client.
+func WithHTTPClientKeyRetrieverOpt(client *http.Client) DefaultKeyRetrieverOption {
```
gamab, line 46: That's a mouthful, but it conflicted with the token exchanger option

authn/jwks.go
```diff
@@ -18,26 +18,42 @@ type KeyRetriever interface {
...
+	}
+
+	for _, o := range opt {
+		o(s)
+	}
+	return s
+}
+
+// WithHTTPClientKeyRetrieverOpt allows setting the HTTP client to be used by the key retriever.
+func WithHTTPClientKeyRetrieverOpt(client *http.Client) DefaultKeyRetrieverOption {
```
kalleep, line 46: Nit: move this to be under the type definition of options

kalleep (approved):

PR #37, opened 2024-05-31
AuthZ: Remove ID Token from the interface

gamab:
~~In scope of the `UniStore` work, I noticed it could be good to share the `IDTokenVerifier` between the authentication and authorization middlewares. The same way we can share a JWKs retriever between `IDTokenVerifier` and `AccessTokenVerifier`.~~

Instead of parsing the id token to extract the `namespacedID` on each authorization check (in other words instead of re-authenticating the user on each check), we'd now request the `namespacedID` to be provided instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ func main() {
 	client, err := authz.NewEnforcementClient(authz.Config{
 		APIURL: "http://localhost:3000",
 		Token:  "<service account token>",
-	}, verifier)
+	})
 
 	if err != nil {
 		log.Fatal("failed to construct authz client", err)
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Grafana needs to be configured with the `accessControlOnCall` feature toggle set
 
 ```ini
 [feature_toggles]
-enable = accessControlOnCall 
+enable = accessControlOnCall
 ```
 
 ## Examples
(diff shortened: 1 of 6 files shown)


kalleep (commented):
TBH I am a bit confused why we use and verifier in the enforcer. Usually you perform authentication and authorization as separate steps, where authentication should happen before. Now authentication is kinda backed into the enforcer but most likley the request has already been authenticated so we parse and verify the claims several times. 

What if we instead changed the interface to take the subject instead of the id token?

gamab:
Yes that's a good point for the `UniStore`, it's true that the request has been authenticated when it reaches the authorization part.

The initial intent of handling the token verification in the enforcer was to allow plugin to do authentication/authorization in a single call. They didn't have to know how the id token was formatted and how to extract the namespacedID from it.

> so we parse and verify the claims several times.

This gets me as well.

Let me work on a change and see how it looks.

README.md
```diff
@@ -12,7 +12,7 @@ Grafana needs to be configured with the `accessControlOnCall` feature toggle set
 
 ```ini
 [feature_toggles]
-enable = accessControlOnCall 
+enable = accessControlOnCall
```
kalleep, line 15: Btw we should remove this requirement for adding the search endpoint. What is stopping us?

gamab: I wanted to give it a bit of time with OnCall `actually` running on prod with it. Since we had a false start there.

gamab:
I have one small worry though. Plugin developers like slo, won't have compile errors if they upgrade the lib and might risk passing the id token still.

README.md
kalleep: I don't mean removing the feature toggle but just always register that endpoint

gamab: But given we now have other plugins relying on it.... I don't think it can be toggled off anyway

authz/models.go
```diff
@@ -5,6 +5,8 @@ import (
 	"net/http"
 )
 
+type NamespacedID string
```
kalleep, line 8: I think we should be careful about adding this. I think we at some point wan't to move out some of these from grafana in here and there we have a typed version.

gamab: yup. Why not start to copy/move the code now?

gamab:
Closing for now, we will create a new client interface instead.

PR #39, opened 2024-06-10
integ unistor

gamab:
** Work in progress to open discussion **


gamab:
This has allowed me to see what I'd need for an authorization client to work with the authorization server. But too many things don't fit nicely together (stackID, token getter, namespacedID instead of jwt, no scope nor action prefix filtering). We will need a multi-tenant version of the client.

PR #40, opened 2024-06-11
Proto files for legacy authz service

gamab:
Adding the authorization service proto files to the auth lib, to be shared between the servers (embedded or mt)


drclau (approved):
🚀

kalleep (commented):
Missing step is to generate the proto files in ci and check if we have a diff

kalleep (approved):

PR #44, opened 2024-07-15
MultiTenantClient: Add caching

gamab:
Add caching to the Legacy Client Impl

### DIFF
--- a/authz/multitenant_client.go
+++ b/authz/multitenant_client.go
@@ -1,10 +1,11 @@
 package authz
 
 import (
+	"bytes"
 	"context"
+	"encoding/gob"
 	"errors"
 	"fmt"
-	"time"
 
 	"github.com/grafana/authlib/authn"
 	authzv1 "github.com/grafana/authlib/authz/proto/v1"
@@ -15,7 +16,6 @@ import (
 
 // TODO (gamab): Instantiate the AuthZ client
 // TODO (gamab): Namespace validation
-// TODO (gamab): Caching
 // TODO (gamab): Logs
 // TODO (gamab): Traces
 // TODO (gamab): AccessToken in outgoing context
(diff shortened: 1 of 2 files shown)


authz/multitenant_client.go
```diff
@@ -225,3 +242,40 @@ func (r *controller) Check(resources ...Resource) bool {
...
+	}
+
+	buf := bytes.Buffer{}
+	err := gob.NewEncoder(&buf).Encode(*ctrl)
+	if err != nil {
+		return err
+	}
+
+	// Cache with default expiry
+	return c.cache.Set(ctx, key, buf.Bytes(), cache.DefaultExpiration)
```
drclau, line 266: At a glance, I think we can eliminate the need to serialize/deserialize, if we modify our cache implementation to take/return a `*controller` instead of a `[]byte`. wdyt?

drclau: I see `Cache` is already used in other parts of the code. It could take `any` instead of `*controller`. In these circumstances it's not going to be less type-safe than using `[]byte`, IMO. We can also use `generics`.

drclau (approved):
LGTM, w/ a couple comments.

authz/multitenant_client.go
gamab: Is there a drawback to switching to any for cache impl were serialization would be required? (Thinking of a Redis Implementation, I assume you have to cache bytes there)

gamab:
We can follow up with your comment in another PR as modifying the cache interface would modify a broader range of files

authz/multitenant_client.go
drclau: Can't think of any drawbacks for using `any`. Since we already have our own internal `Cache` interface, we'd just write new implementations for new caching backends. I would be curious to know why `[]bytes` was chosen to begin with.

In any case, since this change would require updating the already existing usages of `Cache`, I suggest we leave it for another time.

PR #49, opened 2024-07-19
AuthN: Add id token to gRPC client interceptors

gamab:
This PR allows users to add id tokens to the grpc metadata in the outgoing context.

Additionally, to cover future usage, this PR allows adding additional decorators to the metadata.

### DIFF
--- a/authn/grpc_client_interceptors.go
+++ b/authn/grpc_client_interceptors.go
@@ -32,16 +32,16 @@ type GrpcClientConfig struct {
 	// IDTokenMetadataKey is the key used to store the ID token in the outgoing context metadata.
 	// Defaults to "X-Id-Token".
 	IDTokenMetadataKey string
-	// IDTokenExtractor is a function that extracts the ID token from the context.
-	// This is optional and defaults to nil.
-	IDTokenExtractor func(context.Context) (string, error)
 }
 
 // GrpcClientInterceptor is a gRPC client interceptor that adds an access token to the outgoing context metadata.
 type GrpcClientInterceptor struct {
 	cfg         *GrpcClientConfig
 	tokenClient TokenExchanger
+	mdFns       []MetadataDecorator
 }
+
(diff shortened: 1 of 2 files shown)


authn/grpc_client_interceptors.go
```diff
@@ -8,9 +8,11 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-var DefaultAccessTokenMetadataKey = "X-Access-Token"
+var (
```
linoman, line 11: This could perfectly be `const`

authn/grpc_client_interceptors.go
```diff
@@ -32,8 +37,11 @@ type GrpcClientConfig struct {
 type GrpcClientInterceptor struct {
 	cfg         *GrpcClientConfig
 	tokenClient TokenExchanger
+	mdFns       []MetadataDecorator
```
linoman, line 40: Nit: I'm not a fan of abbreviations; it's hard to understand this, even with context.

authn/grpc_client_interceptors.go
```diff
@@ -100,5 +130,12 @@ func (gci *GrpcClientInterceptor) wrapContext(ctx context.Context) (context.Cont
 
 	md.Set(gci.cfg.AccessTokenMetadataKey, token.Token)
 
+	for _, fn := range gci.mdFns {
+		md, err = fn(md)
```
linoman, line 134: I am confused by this code. Why is `md` being assigned to `md` again?

linoman (approved):
Overall, it looks good. I left some nits and questions. It would be great if we could address them before moving on.

Jguer (approved):

Jguer (approved):

authn/grpc_client_interceptors.go
gamab: Accounted for.

PR #50, opened 2024-07-22
AuthZ: Improve tracing for debugging

gamab:
For improved debuggability, we've enhanced tracing capabilities. We realized logging wasn't the best fit for a library.

### DIFF
--- a/authz/multitenant_client.go
+++ b/authz/multitenant_client.go
@@ -175,6 +175,7 @@ func (c *LegacyClientImpl) Check(ctx context.Context, req *CheckRequest) (bool, 
 		return false, err
 	}
 
+	span.SetAttributes(attribute.String("service", req.Caller.AccessTokenClaims.Subject))
 	span.SetAttributes(attribute.Int64("stack_id", req.StackID))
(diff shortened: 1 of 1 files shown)


authz/multitenant_client.go
```diff
@@ -181,6 +195,8 @@ func (c *LegacyClientImpl) Check(ctx context.Context, req *CheckRequest) (bool,
 		return false, nil
 	}
 
+	span.SetAttributes(attribute.String("subject", req.Caller.IDTokenClaims.Subject))
```
drclau, line 198: nit: if you put this line in an `else` block for the `if` above, the safety of dereferencing `IDTokenClaims` will be more explicit, plus it's slightly more future-proof.
Functionally, it's OK as it is too.

drclau (approved):
LGTM, w/ small optional suggestion.

aarongodin (approved):
LGTM, other than Claudiu's existing comment

authz/multitenant_client.go
gamab: The existence of the IDTokenClaims subject is already checked higher up in the validate function, and the subject is used lower down without any check.

PR #51, opened 2024-07-23
AuthZ: Add namespace validation

gamab:
**PR changes:**

- Introduces a configurable namespace formatter, with implementations for onprem (org based namespaces) and cloud (stack based namespaces)
- Adds the namespace validation logic to the AuthZ client. 

**Benefits:**

- **Flexibility:** The authorization client can now easily adapt to different environments with varying namespace formats.
- **Security:**  Access control checks in other namespaces will result in forbidden access.

**Testing:**

- Unit tests have been updated to cover the new namespace formatters.

### DIFF
--- a/authz/multitenant_client.go
+++ b/authz/multitenant_client.go
@@ -186,9 +186,7 @@ func (c *LegacyClientImpl) Check(ctx context.Context, req *CheckRequest) (bool, 
 		return false, err
 	}
 
-	// Validate the namespace
-	ok := c.validateNamespace(req.Caller, req.StackID)
-	if !ok {
+	if !c.validateNamespace(req.Caller, req.StackID) {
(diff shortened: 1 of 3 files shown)


authz/multitenant_client.go
```diff
@@ -175,6 +186,10 @@ func (c *LegacyClientImpl) Check(ctx context.Context, req *CheckRequest) (bool,
 		return false, err
 	}
 
+	if !c.validateNamespace(req.Caller, req.StackID) {
```
gamab, line 189: It feels like this is a bit of authentication logic in the authorization implementation.
But also it does seem to be something you'd want to make sure cannot be bypassed.

drclau (approved):
lgtm

PR #52, opened 2024-07-24
AuthN: gRPC Client Interceptors - Add option to make access tokens optional

gamab:
This PR adds an unsafe option to make access tokens options.
This is meant to be used on-prem while access tokens are not possible to use there.

### DIFF
--- a/authn/grpc_client_interceptors.go
+++ b/authn/grpc_client_interceptors.go
@@ -77,7 +77,7 @@ func NewGrpcClientInterceptor(cfg *GrpcClientConfig, opts ...GrpcClientIntercept
 		gci.cfg.IDTokenMetadataKey = DefaultIdTokenMetadataKey
 	}
 
-	if gci.cfg.TokenRequest == nil {
+	if gci.cfg.TokenRequest == nil && !gci.cfg.DisableAccessToken {
 		return nil, fmt.Errorf("missing required token request: %w", ErrMissingConfig)
 	}
 
@@ -85,7 +85,7 @@ func NewGrpcClientInterceptor(cfg *GrpcClientConfig, opts ...GrpcClientIntercept
 		opt(gci)
 	}
 
-	if gci.tokenClient == nil {
+	if gci.tokenClient == nil && !gci.cfg.DisableAccessToken {
 		if gci.cfg.TokenClientConfig == nil {
 			return nil, fmt.Errorf("missing required token client config: %w", ErrMissingConfig)
 		}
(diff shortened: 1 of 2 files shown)


authn/grpc_client_interceptors.go
```diff
@@ -13,12 +13,11 @@ const (
...
 )
 
-// TODO (gamab): Make Access Token optional?
-// TODO (gamab): Organization/Stack ID
-// TODO (gamab): Readme
-
 // GrpcClientConfig holds the configuration for the gRPC client interceptor.
 type GrpcClientConfig struct {
+	// DisableAccessToken is a flag to disable the access token.
+	// Warning: Using this options means there won't be any servie authentication.
```
dmihai, line 19:
```suggestion
	// Warning: Using this option means there won't be any service authentication.
```

authn/grpc_client_interceptors.go
```diff
@@ -78,15 +77,15 @@ func NewGrpcClientInterceptor(cfg *GrpcClientConfig, opts ...GrpcClientIntercept
...
+	if gci.cfg.TokenRequest == nil && !gci.cfg.DisableAccessToken {
 		return nil, fmt.Errorf("missing required token request: %w", ErrMissingConfig)
 	}
 
 	for _, opt := range opts {
 		opt(gci)
 	}
 
-	if gci.tokenClient == nil {
+	if gci.tokenClient == nil && !gci.cfg.DisableAccessToken {
```
linoman, line 88: Two things

1. I dislike the `true` value for `disabling` something. I find it counterintuitive. I propose a `EnableAccessToken` variable with a default true value.
2. If the Access Token is disabled, then the function `NewGrpcClientInterceptor` will never return the correct `gci`. I think we can improve the code itself by having a single condition for this new variable with a brand new descriptive log that the feature is disabled.

gamab: > I dislike the true value for disabling something. I find it counterintuitive. I propose a EnableAccessToken variable with a default true value.

I put some thoughts into this as I share your concern. But it's not possible to have a default value to true, you would not be able to differentiate unset for set to false. Unless we use a pointer which I'd prefer to avoid.

gamab: > If the Access Token is disabled, the' NewGrpcClientInterceptor' function will never return the correct gci. I think we can improve the code by having a single condition for this new variable with a brand new descriptive log that the feature is disabled.

I don't understand your point. If `Access Token` is disabled then `NewGrpcClientInterceptor` returns a gci that does not add access tokens to the outgoing metadata.

Jguer (approved):

dmihai (approved):

authn/grpc_client_interceptors.go
linoman: @gamab You are correct. I misread the closing conditional statement before the `gci.tokenClient`.

LGTM

PR #54, opened 2024-07-24
AuthZ: LegacyClient - Option to override the gRPC connection

gamab:
I had forgotten the use case where we'd want to run the client with in-proc gRPC.

With this additional option, we can:
```go
channel := &inprocgrpc.Channel{}
channel.RegisterService(
	grpchan.InterceptServer(
		&authzv1.AuthzService_ServiceDesc,
		srvOpts...,
	),
	server,
)

conn := grpchan.InterceptClientConn(channel, clientOpts...)
client, err := NewLegacyClient(
    &MultiTenantClientConfig{},
    WithGrpcConnectionLCOption(conn),
)
```

### DIFF
--- a/authz/multitenant_client.go
+++ b/authz/multitenant_client.go
@@ -57,6 +57,7 @@ type LegacyClientImpl struct {
 	authCfg      *MultiTenantClientConfig
 	clientV1     authzv1.AuthzServiceClient
 	cache        cache.Cache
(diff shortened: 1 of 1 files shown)


Jguer (approved):

gamab:
I know it's not perfect, we might have to iterate to find the right instantiation options. :pray:

PR #56, opened 2024-07-25
Authn: Add DisplayName to ID Token Claims

colin-stuart:
Related to https://github.com/grafana/grafana/pull/90992

Adds `DisplayName` as a field called `name` to the ID Token claims.

### DIFF
--- a/authn/verifier_id_token.go
+++ b/authn/verifier_id_token.go
@@ -16,6 +16,8 @@ type IDTokenClaims struct {
 	Username string `json:"username"`
 	// UID is the unique ID of the user (UID attribute)
 	UID string `json:"uid"`
+	// Name of the user (name attribute)
+	Name string `json:"name"`
 }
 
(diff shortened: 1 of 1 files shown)


authn/verifier_id_token.go
```diff
@@ -16,6 +16,8 @@ type IDTokenClaims struct {
 	Username string `json:"username"`
 	// UID is the unique ID of the user (UID attribute)
 	UID string `json:"uid"`
+	// Display Name of the user (name attribute if it is set, otherwise the login or email)
+	DisplayName string `json:"name"`
```
mgyongyosi, line 20: nit: is there any specific reason to have `name` and not `displayName` as the json attribute?

mgyongyosi (approved):

authn/verifier_id_token.go
colin-stuart: I thought it would be simpler for consumers of the ID token to refer to this attribute as `name`, but it could be changed to `displayName` for consistency.

PR #57, opened 2024-07-26
AuthN: Make gRPC authenticator configurable

gamab:
This PR makes the gRPC authenticator configurable.

It adds a constructor to the gRPC Authenticator with configuration and options.

One of the option is to make id tokens optional or required.

### DIFF
--- a/authn/grpc_authenticator_test.go
+++ b/authn/grpc_authenticator_test.go
@@ -24,10 +24,12 @@ func setupGrpcAuthenticator() *testEnv {
 		idVerifier: &fakeIDTokenVerifier{},
 	}
 	env.authenticator = &GrpcAuthenticator{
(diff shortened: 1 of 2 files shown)


Jguer (approved):

authn/grpc_authenticator_test.go
```diff
@@ -24,14 +24,60 @@ func setupGrpcAuthenticator() *testEnv {
...
 	}
+	setDefaultMetadataKeys(env.authenticator.cfg)
 
 	return env
 }
 
+func TestGrpcAuthenticator_NewGrpcAuthenticator(t *testing.T) {
+	t.Run("should return error when missing signing keys URL", func(t *testing.T) {
+		ga, err := NewGrpcAuthenticator(&GrpcAuthenticatorConfig{})
+		require.Error(t, err)
```
linoman, line 40:
```suggestion
```

I believe this is not needed when the following line is required.

linoman (approved):
LGTM, I left a nit

PR #58, opened 2024-07-26
AuthN: gRPC authenticator option to disable access tokens

gamab:
This PR adds an unsafe option to make access tokens optionals.
This is meant to be used on-prem while access tokens are not possible to use there.

### DIFF
--- a/authn/grpc_authenticator.go
+++ b/authn/grpc_authenticator.go
@@ -79,7 +79,7 @@ func WithIDTokenAuthOption(required bool) GrpcAuthenticatorOption {
 	}
 }
 
-// WithDisableAccessTokenAuthOption is a flag to disable access token authentication.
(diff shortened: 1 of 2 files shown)


authn/grpc_authenticator.go
```diff
@@ -44,6 +43,10 @@ type GrpcAuthenticatorConfig struct {
 	// VerifierConfig holds the configuration for the token verifiers.
 	VerifierConfig VerifierConfig
 
+	// accessTokenAuthEnabled is a flag to enable access token authentication.
+	// If disabled, only ID token authentication is performed.
+	// Warning: Using this option means there won't be any service authentication.
+	accessTokenAuthEnabled bool
```
gamab, line 49: @linoman figured out a way to have an `enabled` instead of `disabled` logic in the end. :tada:

linoman (approved):

PR #60, opened 2024-07-29
PoC AuthN: Authenticator can validate token namespaces based on the request

gamab:
In this PR, I'm trying to see what it would take to do the namespace validation based on a stack ID that could be taken from the request, the metadata or the tokens.
It looks complicated to handle the gRPC stream scenario with that approach. 

image

An alternative would be not to do any namespace validation in this middleware and let the handler code or another middleware do it.

This is what I'm assessing in #61


gamab:
Favored #61

PR #61, opened 2024-07-29
AuthNZ: Extract namespace validation from gRPC authentication 

gamab:
**Aka: three step authnz (authentication, namespace access control, permission access control)**

In #60 I tried to assess if I could extract the stackID required for the namespace check from different sources. It turns out it's complicated in the gRPC Stream case.
It made me question whether performing namespace validation in the authentication interceptor was the correct thing after all.

- Authentication: Verifying the identity of the caller. In our case, this is handled by validating the JWT tokens against the signing authority. We've already confirmed **who** is making the request.
- Authorization: Determining if the authenticated caller has the necessary permissions to access a specific resource or perform a specific action. This is where the **namespace validation** and **rbac permission checks** comes into play.

With this PR I'm extracting namespace validation from the gRPC authentication code. The namespace validation code can now be placed in its own interceptor or called at handler level; in the case where the stackID is part of the message which is the case for the authz-service, that makes sense.

### DIFF
--- a/authn/grpc_authenticator.go
+++ b/authn/grpc_authenticator.go
@@ -3,7 +3,6 @@ package authn
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"strings"
 
 	"google.golang.org/grpc/codes"
@@ -33,9 +32,6 @@ type GrpcAuthenticatorConfig struct {
 	// IDTokenMetadataKey is the key used to retrieve the ID token from the incoming metadata.
 	// Defaults to "X-Id-Token".
 	IDTokenMetadataKey string
-	// StackIDMetadataKey is the key used to retrieve the stack ID from the incoming metadata.
-	// Defaults to "X-Stack-Id".
-	StackIDMetadataKey string
 
 	// KeyRetrieverConfig holds the configuration for the key retriever.
 	// Ignored if KeyRetrieverOption is provided.
@@ -60,7 +56,6 @@ type GrpcAuthenticator struct {
 	keyRetriever KeyRetriever
 	atVerifier   Verifier[AccessTokenClaims]
 	idVerifier   Verifier[IDTokenClaims]
-	namespaceFmt NamespaceFormatter
 }
 
 func WithKeyRetrieverOption(kr KeyRetriever) GrpcAuthenticatorOption {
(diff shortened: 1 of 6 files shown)


authn/grpc_authenticator.go
```diff
@@ -15,12 +14,11 @@ var (
 	ErrorMissingMetadata    = status.Error(codes.Unauthenticated, "unauthenticated: no metadata found")
 	ErrorMissingIDToken     = status.Error(codes.Unauthenticated, "unauthenticated: missing id token")
 	ErrorMissingAccessToken = status.Error(codes.Unauthenticated, "unauthenticated: missing access token")
-	ErrorInvalidStackID     = status.Error(codes.PermissionDenied, "unauthorized: invalid stack ID")
 	ErrorInvalidIDToken     = status.Error(codes.PermissionDenied, "unauthorized: invalid id token")
 	ErrorInvalidAccessToken = status.Error(codes.PermissionDenied, "unauthorized: invalid access token")
 )
 
-// TODO (gamab) - StackID extract should be configurable - could come from the metadata, path, id token.
+// TODO (gamab) - Validate service and user namespace match ?
```
gamab, line 21: In othe words: Should we validate here that the service namespace includes the user namespace?

authn/grpc_authenticator.go
```diff
@@ -136,36 +126,33 @@ func (ga *GrpcAuthenticator) Authenticate(ctx context.Context) (context.Context,
...
-		idClaims, err := ga.authenticateUser(ctx, stackIDInt, md)
+		idClaims, err := ga.authenticateUser(ctx, md)
 		if err != nil {
 			return nil, err
 		}
 		callerInfo.IDTokenClaims = idClaims
 	}
 
+	// Validate accessToken namespace matches IDToken namespace
+	if ga.cfg.accessTokenAuthEnabled && ga.cfg.idTokenAuthEnabled && callerInfo.IDTokenClaims != nil {
```
gamab, line 146: It does not hurt to check that at least the service and user tokens are in the same namespace. (Even if I removed verifying the caller is authorized in the namespace)

drclau (approved):

PR #63, opened 2024-08-01
Expose claims as a zero dependency interface

ryantxu:
This PR explores how we could expose claims as a simple interface rather than one that depends directly on `go-jose`

Moving forward, [claims.AuthInfo](https://github.com/grafana/authlib/compare/with-claims-interface?expand=1#diff-b48fcc141a5aa68fd5959d96507c79dd13c1f45324dc5ed0ae69b0fddf29e5f1R5) could be the interface we use as often as possible.

identity.Requester could extend/implement AuthInfo and over time we move as much as possible away from requester/SignedInUser/Identity/user.Info


claims/client.go
ryantxu, line 1: This is added mostly for discussion -- I think something close to this covers everything needed for app platform + unified storage.

kalleep (commented):
Nice, I agree that using interface is a much better approach and is something I should have started with.. Overall looks really good but have a couple of questions / suggestions

claims/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/grafana/authlib/claims
```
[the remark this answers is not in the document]
ryantxu: With a separate module and limited dependencies -- I feel much better about promoting this as a dependency *everywhere* -- obviously the host processes will need to implement the claims + client somewhere.

Essentially this lets us add authlib/claims as a dependency to every plugin/library without also depending on go-jose and everything else in https://github.com/grafana/authlib/blob/main/go.mod#L8

claims/client.go
```diff
@@ -0,0 +1,45 @@
+package claims
+
+import "context"
+
+// AccessRequest describes the requested access.
+// This is designed bo to play nicely with the kubernetes authorization system:
+// https://github.com/kubernetes/kubernetes/blob/v1.30.3/staging/src/k8s.io/apiserver/pkg/authorization/authorizer/interfaces.go#L28
+type AccessRequest struct {
```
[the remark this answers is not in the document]
ryantxu: I don't think so -- we can have an Authorizer that calls authlib

claims/token.go
```diff
@@ -0,0 +1,55 @@
...
+type TokenClaims interface {
+	Issuer() string
+	Subject() string
+	Audience() []string
+	Expiry() *time.Time
+	NotBefore() *time.Time
+	IssuedAt() *time.Time
+
+	// ID for the claims (not the identity!)
+	ID() string
```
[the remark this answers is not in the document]
ryantxu: good call -- I changed it to "JTI()" and added comments from https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7

authn/claims.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+type AuthInfo struct {
+	IdentityClaims *Identity
+	AccessClaims   *Access
+}
+
+type Identity struct {
+	claims Claims[IDTokenClaims]
+}
```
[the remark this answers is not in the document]
ryantxu: ugg -- no great way to define the interface with generics and avoid collisions.  If we use `type Identity Claims[IDTokenClaims]` then the `*jwt.Claims.Audience` property collides with the function `claims.Audience()`

> but then we would break the api.
Once we have the claims module exposed/stable/usable -- then I think we can consider breaking the internal representations so they play nicer; but we don't have to do that at this stage

claims/verifier.go
ryantxu, line 1: Added this for discussion... I'm not sure how I feel about it

claims/verifier.go
```diff
@@ -0,0 +1,9 @@
+package claims
+
+import "context"
+
+type Verifier[T any] interface {
+	// Verify will parse and verify provided token, if `AllowedAudiences` was configured those will be validated as well.
+	// Additional claims will be
+	VerifyToken(ctx context.Context, token string) (TokenClaims, T, error)
+}
```
ryantxu, line 9: Or we could do something more like the [upstream implementation](https://github.com/grafana/authlib/blob/main/authn/verifier.go#L61)
```suggestion
type Verifier interface {
	// Verify will parse and verify provided token, if `AllowedAudiences` was configured those will be validated as well.
	VerifyToken(ctx context.Context, token string, claims any) (TokenClaims, error)
}
```

authn/claims.go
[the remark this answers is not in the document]
kalleep: It works but you have to use the full path for properties e.g.
```go
func (i *Identity) Audience() string {
  return i.Claims.Audience
}
```
The thing I don't like with this approach is that properties are still accessible directly. With what we have now they are not exposed so we should probably keep what we have now

claims/token.go
```diff
@@ -0,0 +1,125 @@
...
+	// [RFC 7519 §4.1.7]: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7
+	JTI() string
+}
+
+type IdentityClaims interface {
+	TokenClaims
+
+	// UID is the unique ID of the user (UID attribute)
+	// This will often be the string version of a TypedID `<type>:<id>`
+	UID() string
```
kalleep, line 88: Should we return the typed id here instead of string encoded version? I started to convert to using that inside core grafana instead

kalleep (commented):
Nice, just one comment left and after that we should be good to merge

claims/token.go
kalleep: hm just saw https://github.com/grafana/grafana/pull/90937. Now we really have an awkward mix of typed and string encoded version. I think we should settle with one of them and stick to that.

There are pros and cons for both of them. It is sometimes really annoying to work with the string version but you could probably hide that away somehow. I noticed https://github.com/grafana/grafana/pull/90937/files#diff-a6d40da7a17d43622287c2147606ef055a394f2110ac283dea3c10d5fe64f216R84 for the typed version you could just call `GetUID().String()`. WDYT?

ryantxu: To play [nice with k8s](https://github.com/grafana/grafana/pull/90937), I think we should stick with UID==string

>  It is sometimes really annoying to work with the string version
what case?

I think we could add `Type()` property to the identity claims and read it from the UID -- or make people do it explicitly 🤷🏻

drclau (approved):
I like this. I'm approving already, and will assume you and Kalle will sort out the remaining open comments.

I am not entirely sure how we can match the `AccessClient / AccessRequest` with RBAC, but we need to cross that bridge very soon and I think it's good to have this code in already.

claims/token.go
kalleep: Agreed, I will work on removing the typed variant so we only stick with the string version

kalleep (approved):
Nice, great work. I wonder if we should have a constructor for TypeIDs in this package
```go
func NewTypedID(typ IdentityType, id string) string {
  return fmt.Sprintf("%s, %s", typ, id)
}
```
It is just convenient to have a function that creates one with the correct format. Also this is something we can add in another pr

PR #65, opened 2024-08-13
Adjust ID Token claims

kalleep:
After inspecting what we get in our claims we need to adjust the methods a bit, specifically UID is sent with a prefix.
I will also make a pr in grafana to add the identity type to all issued tokens.

We also probably won't need InternalID and OrgID for now so lets remove them

Edit:
After talk with Ryan we have adjusted the interface and id claims a bit. Nothing should be relying on these types yet so we should be fine to break them

### DIFF
--- a/authn/verifier_id_token.go
+++ b/authn/verifier_id_token.go
@@ -22,7 +22,7 @@ type IDTokenClaims struct {
 	// Username of the user (login attribute on the Identity)
 	Username string `json:"username,omitempty"`
 	// Display name of the user (name attribute if it is set, otherwise the login or email)
-	DisplayName string `json:"name"`
+	DisplayName string `json:"name,omitempty"`
 }
 
 // Helper for the id
--- a/authn/verifier_id_token.go
+++ b/authn/verifier_id_token.go
@@ -8,7 +8,7 @@ import (
 )
 
 type IDTokenClaims struct {
-	// UID is the unique ID of the user (UID attribute)
+	// UID is the unique ID of the of entity, it is a typed id e.g. `user:some-uid`.
 	UID string `json:"uid"`
 	// The type of user
 	Type claims.IdentityType `json:"type"`
@@ -40,7 +40,7 @@ func (c IDTokenClaims) getK8sName() string {
 	if c.Email != "" {
 		return c.Email
 	}
-	return c.asTypedUID()
+	return c.UID
 }
 
 func (c IDTokenClaims) NamespaceMatches(namespace string) bool {
--- a/authn/verifier_id_token.go
+++ b/authn/verifier_id_token.go
@@ -20,7 +20,7 @@ type IDTokenClaims struct {
 	Email           string `json:"email,omitempty"`
 	EmailVerified   bool   `json:"email_verified,omitempty"`
 	// Username of the user (login attribute on the Identity)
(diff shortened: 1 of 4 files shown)


ryantxu (approved):

Jguer (approved):

mgyongyosi (approved):

authn/verifier_id_token.go
```diff
@@ -8,16 +8,10 @@ import (
 )
 
 type IDTokenClaims struct {
-	// UID is the unique ID of the user (UID attribute)
+	// UID is the unique ID of the of entity, it is a typed id e.g. `user:some-uid`.
 	UID string `json:"uid"`
 	// The type of user
 	Type claims.IdentityType `json:"type"`
```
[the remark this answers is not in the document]
kalleep: I am not too found of it but it can be a hassle to remove the prefix without breaking anything.

Right now we set:
* Subject - `<type>:<internal id>`
* UID - `<type>:<uid>`

On the interface we have two methods:
*  `UID()`  returns the uid without type prefix
* `GetUID()` returns uid with prefix

I kept the IdentityType in claims because it is needed for `IdentityType` from interface. We could remove it but then we need to parse it out from either subject or uid field before returning the claims.

authn/verifier_id_token.go
```diff
@@ -8,32 +8,26 @@ import (
...
-	InternalID int64 `json:"id,omitempty"`
-	// The internal numeric org ID
-	// Deprecated: use namespace where possible
-	OrgID int64 `json:"orgId,omitempty"`
 	// Namespace takes the form of '<type>-<id>', '*' means all namespaces.
 	// Type can be either org or stack.
 	Namespace string `json:"namespace"`
 	// AuthenticatedBy is the method used to authenticate the identity.
 	AuthenticatedBy string `json:"authenticatedBy"`
 	Email           string `json:"email"`
```
ryantxu, line 20:
```suggestion
	AuthenticatedBy string `json:"authenticatedBy,omitempty"`
	Email           string `json:"email,omitempty"`
```

ryantxu, line 20:
```suggestion
	AuthenticatedBy string `json:"authenticatedBy,omitempty"`
	Email           string `json:"email,omitempty"`
```

authn/verifier_id_token.go
```diff
@@ -8,32 +8,26 @@ import (
...
-	EmailVerified   bool   `json:"email_verified"`
+	EmailVerified   bool   `json:"email_verified,omitempty"`
 	// Username of the user (login attribute on the Identity)
 	Username string `json:"username"`
```
ryantxu, line 23:
```suggestion
	Username string `json:"username,omitempty"`
```

authn/verifier_id_token.go
```diff
@@ -8,32 +8,26 @@ import (
...
-	// Display Name of the user (name attribute if it is set, otherwise the login or email)
+	// Display name of the user (name attribute if it is set, otherwise the login or email)
 	DisplayName string `json:"name"`
```
ryantxu, line 25:
```suggestion
	DisplayName string `json:"name,omitempty"`
```

drclau (approved):

PR #67, opened 2024-08-14
Add lint step to ci

kalleep:
The main objective with this pr is to add a lint step to our ci pipeline.

When running this on our repository we get a lot of deprecation issues. Some are trivial to fix e.g. `NamespaceFormatter`.

In https://github.com/grafana/authlib/pull/63 we moved them to claims package but did not replace internal usage. This will break existing code but should be easy to fix by just importing the correct formatter from claims package instead.

There are other changes we probably need to do such as replacing usage of `authn.CallerAuthInfo` inside our authz client code with the interface `claims.AuthInfo`. Those are a bit bigger changes and I prefer to do them in a separate pr.


authn/caller_info.go
```diff
@@ -10,7 +10,7 @@ var (
 	_ claims.AuthInfo = &CallerAuthInfo{}
 )
 
-// Deprecated: Use claims.AuthInfo
+// Deprecated: Use authn.AuthInfo
```
kalleep, line 13: @ryantxu sorry for the ping but we should use authn.AuthInfo instead of authn.CallerAuthInfo right, at least when it comes to implementation

IevaVasiljeva (approved):
LGTM

drclau:
> There are other changes we probably need to do such as replacing usage of authn.CallerAuthInfo inside our authz client code with the interface claims.AuthInfo. Those are a bit bigger changes and I prefer to do them in a separate pr.

@kalleep, I will put a PR for `authn.CallerAuthInfo` -> `claims.AuthInfo` shortly.

drclau (approved):

PR #68, opened 2024-08-14
Switching from `authn.CallerAuthInfo` to `claims.AuthInfo`

drclau:
Follow-up to https://github.com/grafana/authlib/pull/63

`authn.CallerAuthInfo` is being deprecated and replaced by `claims.AuthInfo`.

### DIFF
--- a/authn/grpc_authenticator.go
+++ b/authn/grpc_authenticator.go
@@ -8,6 +8,8 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
+
+	"github.com/grafana/authlib/claims"
 )
 
 var (
@@ -119,7 +121,7 @@ func NewGrpcAuthenticator(cfg *GrpcAuthenticatorConfig, opts ...GrpcAuthenticato
 
 // Authenticate authenticates the incoming request based on the access token and ID token, and returns the context with the caller information.
 func (ga *GrpcAuthenticator) Authenticate(ctx context.Context) (context.Context, error) {
-	callerInfo := CallerAuthInfo{}
+	authInfo := AuthInfo{}
 
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
@@ -131,7 +133,9 @@ func (ga *GrpcAuthenticator) Authenticate(ctx context.Context) (context.Context,
 		if err != nil {
 			return nil, err
 		}
-		callerInfo.AccessTokenClaims = *atClaims
+		if atClaims != nil {
+			authInfo.AccessClaims = &Access{claims: *atClaims}
+		}
 	}
 
 	if ga.cfg.idTokenAuthEnabled {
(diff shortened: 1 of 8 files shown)


authz/multitenant_client.go
```diff
@@ -272,13 +280,18 @@ func (c *LegacyClientImpl) Check(ctx context.Context, req *CheckRequest) (bool,
...
-	accessTokenMatch := !c.authCfg.accessTokenAuthEnabled || caller.AccessTokenClaims.Rest.NamespaceMatches(expectedNamespace)
-	idTokenMatch := caller.IDTokenClaims == nil || caller.IDTokenClaims.Rest.NamespaceMatches(expectedNamespace)
+	accessClaims := caller.GetAccess()
+	accessTokenMatch := !c.authCfg.accessTokenAuthEnabled ||
+		(accessClaims != nil && !accessClaims.IsNil() &&
+			(accessClaims.Namespace() == "*" || accessClaims.Namespace() == expectedNamespace))
+
+	idClaims := caller.GetIdentity()
+	idTokenMatch := idClaims == nil || idClaims.IsNil() ||
+		idClaims.Namespace() == "*" || idClaims.Namespace() == expectedNamespace
```
kalleep, line 294: I think we should add NamespacesMatches to both `claims.AccessClaims` and `claims.IdentityClaims`, internally it can just be delegated to the functions we have for this

authn/grpc_authenticator.go
```diff
@@ -131,25 +133,37 @@ func (ga *GrpcAuthenticator) Authenticate(ctx context.Context) (context.Context,
...
+		}
 	}
 
 	if ga.cfg.idTokenAuthEnabled {
 		idClaims, err := ga.authenticateUser(ctx, md)
 		if err != nil {
 			return nil, err
 		}
-		callerInfo.IDTokenClaims = idClaims
+		if idClaims != nil {
```
kalleep, line 146: is this path possible, I would expect either we always get an error or valid claims

authn/grpc_authenticator.go
```diff
@@ -131,25 +133,37 @@ func (ga *GrpcAuthenticator) Authenticate(ctx context.Context) (context.Context,
 		if err != nil {
 			return nil, err
 		}
-		callerInfo.AccessTokenClaims = *atClaims
+		if atClaims != nil {
```
kalleep, line 136: is this path possible, I would expect either we always get an error or valid claims

authn/grpc_authenticator.go
drclau: We can get in this situation if ID tokens are enabled but not required, because of [this code](https://github.com/grafana/authlib/blob/e71d90e439c7ae19ec3a5ed7b3a1117f41dfafaa/authn/grpc_authenticator.go#L193-L199).
There's also a [test for this specific scenario](https://github.com/grafana/authlib/blob/e71d90e439c7ae19ec3a5ed7b3a1117f41dfafaa/authn/grpc_authenticator_test.go#L139-L156), which panics without the check.

For the other place (`atClaims`), I don't think we can get in the `atClaims == nil` situation right now, but I'm inclined to keep the check as future proofing. I'm OK to remove it, too.

kalleep: Ah yeah I did not see that is was different options

kalleep (approved):
Nice, after you have merged the pr you need to do an additional pr to sync the new version for claims. It is a bit annoying :/

PR #70, opened 2024-08-15
Pass authz's http client to verifier

mahendrapaipuri:
The PR passes the HTTP client of `authz` to the `KeyRetriever`. This ensures any options passed to `authz` client will be passed to `KeyRetriever`. A common use case is to skip TLS checks in development environment where this option must be passed to `KeyRetriever` as well to ignore certificate errors while fetching signing keys.

PR #72, opened 2024-08-15
Fix token type verification

### LINKED ISSUE #71: `authz`'s `EnforcementClient` does not work with Grafana `< 11.1.0`
`authzClient.HasAccess` does not work using the [latest]() commit on `main` branch using Grafana `< 11.1.0`. We get following error:

```
[grafana] | logger=plugin.grafana-appwithrbac-app t=2024-08-15T15:57:04.80338844Z level=error msg="Error checking access" endpoint=callResource error="invalid id token: cannot extract namespaced ID: invalid token: invalid token type" pluginID=grafana-appwithrbac-app
```

**Reproducer:**

- Use the plugins examples [app-with-rbac](https://github.com/grafana/grafana-plugin-examples/tree/main/examples/app-with-rbac).
- Update `authlib` to latest commit `go get -u github.com/grafana/authlib@2f8d13a83ca3e82da08b53726de1697ee5b5b4cc && go mod tidy`
- Run compose using Grafana 10.4.0 `GRAFANA_VERSION=10.4.0 docker-compose up --build`
- Build plugin using `npm install && npm build && mage -v build:linux`
- Login into Grafana running at `http://localhost:3000` with default admin credentials and go to `http://localhost:3000/a/grafana-appwithrbac-app/patents?`

If we use latest Grafana `11.1.0`, it works. Maybe the changes made in the lib are not backward compatible?

mahendrapaipuri:
Fix #71

### DIFF
--- a/authn/config.go
+++ b/authn/config.go
@@ -9,11 +9,16 @@ import (
 
 type VerifierConfig struct {
 	AllowedAudiences jwt.Audience `yaml:"allowedAudiences"`
+	// Disable token typ header check when it is not present
+	DisableTypHeaderCheck bool `yaml:"disableTypHeaderCheck"`
 }
 
 func (c *VerifierConfig) RegisterFlags(prefix string, fs *flag.FlagSet) {
+	fs.BoolVar(&c.DisableTypHeaderCheck, prefix+".disable-typ-header-check", false, "Disable typ header check in JWT token")
+
 	fs.Func(prefix+".allowed-audiences", "Specifies a comma-separated list of allowed audiences.", func(v string) error {
 		c.AllowedAudiences = jwt.Audience(strings.Split(v, ","))
+
 		return nil
 	})
 }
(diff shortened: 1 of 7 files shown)


kalleep:
Hey, thanks for the pr. We do not want to make typ header optional. To solve the issue it would be better with a configuration option to disable the check. So this have to be an opt out.

mahendrapaipuri:
@kalleep I have pushed few changes:

* Added a new client option `WithVerifier` to the enforcement client. It lets users to configure a verifier, if they need. If it is not configured, a default verfiier will be created as before. 
* Add a new config opt to verifier `DisableTypHeaderCheck` that defaults to `false` if not configured. 

These two changes keep the clients' behaviour as before, so no breaking changes, albeit, it gives more configurability for end users. This changes addresses #70 too as we can directly configure verifier on the enforcement client, so users can configure the same HTTP client that they use for enforcement.

Is it something acceptable?

kalleep (commented):
Nice, this looks better. I also want @drclau to take a look at the pr before I approve it

kalleep:
Hey there, thanks for your patience while we discussed this PR internally. We appreciate you taking the time to contribute!
We've decided not to incorporate this change into the library at this time. Since the library is still in early development and this functionality depends on an experimental feature toggle for Grafana versions earlier than 11.3.x, we want to force the type header check for now. This means the current version won't work as intended until Grafana is updated to a compatible version.

PR #73, opened 2024-08-22
NamespaceMapper: conform to cloud pluralized stacks-<id> namespace

charandas:
There has been some miscommunication within our working group and it made its way into OSS code and into auth/authlib, which is that cloud namespace format is `stack-<stackID>`.

This is incorrect. Cloud namespaces are `stacks-<stackId>` historically.

[Slack convo](https://raintank-corp.slack.com/archives/C037H5GAY9L/p1724257160817399?thread_ts=1722433605.876979&cid=C037H5GAY9L).

This change-set tries to take care of the problem in a backwards-compatible way. In talking with @kalleep, we decided to treat both `stack-<stackId>` and `stacks-<stackId>` as valid for the time being.

While I was at it, I was very happy to see that something like NamespaceAccessChecker exists. I would love to use it for some use-cases but its reliance on me knowing `stackID` was something discouraging due to the chicken-in-egg issue arising from the plural vs singular stack namespaces.

I changed the interface slightly to just take a string `expectedNamespace` and moved the older contract to the more general `CheckAccessForIdentitfier(caller, identifier)`, so it could be used for stackId or orgId.

Very open to feedback since this library is not under my purview. Thanks for considering any changes that you may from the PR.

### DIFF
--- a/authz/namespace.go
+++ b/authz/namespace.go
@@ -3,6 +3,7 @@ package authz
 import (
 	"context"
 	"strconv"
+	"strings"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -27,8 +28,11 @@ var (
 )
 
 type NamespaceAccessChecker interface {
-	CheckAccess(caller claims.AuthInfo, stackID int64) error
+	CheckAccess(caller claims.AuthInfo, namespace string) error
+	CheckAccessForIdentitfier(caller claims.AuthInfo, id int64) error
 }
+
+var _ NamespaceAccessChecker = &NamespaceAccessCheckerImpl{}
 
 type NamespaceAccessCheckerOption func(*NamespaceAccessCheckerImpl)
 
@@ -79,15 +83,14 @@ func NewNamespaceAccessChecker(namespaceFmt claims.NamespaceFormatter, opts ...N
 	return na
 }
 
-func (na *NamespaceAccessCheckerImpl) CheckAccess(caller claims.AuthInfo, stackID int64) error {
-	expectedNamespace := na.namespaceFmt(stackID)
+func (na *NamespaceAccessCheckerImpl) CheckAccess(caller claims.AuthInfo, expectedNamespace string) error {
 	if na.idTokenEnabled {
 		idClaims := caller.GetIdentity()
 		if idClaims == nil || idClaims.IsNil() {
 			if na.idTokenRequired {
 				return ErrorMissingIDToken
 			}
-		} else if idClaims.Namespace() != expectedNamespace {
+		} else if !checkEqualsNamespaceDisambiguous(expectedNamespace, idClaims.Namespace()) {
 			return ErrorIDTokenNamespaceMismatch
 		}
 	}
@@ -97,11 +100,18 @@ func (na *NamespaceAccessCheckerImpl) CheckAccess(caller claims.AuthInfo, stackI
 			return ErrorMissingAccessToken
 		}
 		namespace := accessClaims.Namespace()
-		if namespace != "*" && namespace != expectedNamespace {
+		if namespace != "*" && !checkEqualsNamespaceDisambiguous(expectedNamespace, namespace) {
 			return ErrorAccessTokenNamespaceMismatch
 		}
 	}
 	return nil
+}
+
+// CheckAccessForIdentitfier needs an identifier which it uses the preconfigured namespace formatter
+// to generate the expected namespace from.
+func (na *NamespaceAccessCheckerImpl) CheckAccessForIdentitfier(caller claims.AuthInfo, id int64) error {
+	expectedNamespace := na.namespaceFmt(id)
+	return na.CheckAccess(caller, expectedNamespace)
 }
 
 type StackIDExtractors func(context.Context) (int64, error)
@@ -140,7 +150,7 @@ func UnaryNamespaceAccessInterceptor(na NamespaceAccessChecker, stackID StackIDE
 			return nil, err
 		}
 
-		err = na.CheckAccess(caller, stackID)
+		err = na.CheckAccessForIdentitfier(caller, stackID)
 		if err != nil {
 			return nil, err
 		}
@@ -164,7 +174,7 @@ func StreamNamespaceAccessInterceptor(na NamespaceAccessChecker, stackID StackID
 			return err
 		}
 
-		err = na.CheckAccess(caller, stackID)
+		err = na.CheckAccessForIdentitfier(caller, stackID)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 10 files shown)


authz/namespace.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	DefaultStackIDMetadataKey = "X-Stack-ID"
+	DefaultStackIDMetadataKey = "X-stacks-ID"
```
toddtreece, line 16: does this need to change?

kalleep (changes requested):
Okey, this requires a bit more work. We need to update in at least one other place before we merge this one so lets hold this one until that work is done

authz/namespace.go
kalleep: Probably result of search and replace

charandas: Will revert

authz/namespace.go
```diff
@@ -26,15 +28,26 @@ var (
...
 )
 
+type NamespaceAccessCheckerType int
+
+const (
+	NamespaceAccessCheckerTypeCloud NamespaceAccessCheckerType = iota + 1
+	NamespaceAccessCheckerTypeOrg
+)
+
 type NamespaceAccessChecker interface {
```
charandas, line 38: I am thinking of using this interface in both `ext_jwt` in OSS and in other places in enterprise. But its original contract is not gonna help with resolving the ambiguity, so I moved that to `CheckAccessForIdentitifer` (`Identifier` instead of `StackID` to be more general).

charandas: Here are the associated OSS changes: https://github.com/grafana/grafana/pull/92332/commits/b5807e210a0efecc522170a43b5bb86679bccca6

charandas:
[L90](https://github.com/grafana/authlib/blob/main/authz/namespace.go#L90C13-L90C54) in current main won't work, as service layer when acting on behalf of the user may call `namespaceAccessChecker.Check(requester, "*")` but since id tokens never have such a claim, L90 would fail the check, when it should actually let it pass. So I will add some check for that.

authz/namespace.go
```diff
@@ -184,3 +218,27 @@ func getFirstMetadataValue(md metadata.MD, key string) (string, bool) {
...
+	if len(actualNamespaceParts) < 2 {
+		return false
+	}
+
+	expectedNamespaceParts := strings.Split(expectedNamespace, "-")
+	if len(expectedNamespaceParts) < 2 {
+		return false
+	}
+
+	if checkerType == NamespaceAccessCheckerTypeCloud && (actualNamespaceParts[0] == "stack" || actualNamespaceParts[0] == "stacks") && expectedNamespaceParts[0] == "stacks" {
```
charandas, line 239: Since `expectedNamespace` is generated by the library, it doesn't need the singular case. Maybe not very intuitive, but efficient!

claims/namespace.go
```diff
@@ -53,13 +53,13 @@ func ParseNamespace(ns string) (NamespaceInfo, error) {
 		return info, err
 	}
 
-	if strings.HasPrefix(ns, "stack-") {
+	if strings.HasPrefix(ns, "stacks-") {
 		stackIDStr := ns[6:]
-		stackID, err := strconv.ParseInt(stackIDStr, 10, 64)
-		if err != nil || stackID < 1 {
-			return info, fmt.Errorf("invalid stack id")
+		stackId, err := strconv.Atoi(stackIDStr)
```
gamab, line 58: Why this change?

claims/namespace.go
```diff
@@ -53,13 +53,13 @@ func ParseNamespace(ns string) (NamespaceInfo, error) {
 		return info, err
 	}
 
-	if strings.HasPrefix(ns, "stack-") {
+	if strings.HasPrefix(ns, "stacks-") {
 		stackIDStr := ns[6:]
```
gamab, line 57:
```suggestion
	if strings.HasPrefix(ns, "stacks-") {
		stackIDStr := ns[7:]
```

claims/namespace.go
```diff
@@ -53,13 +53,13 @@ func ParseNamespace(ns string) (NamespaceInfo, error) {
 		return info, err
 	}
 
-	if strings.HasPrefix(ns, "stack-") {
+	if strings.HasPrefix(ns, "stacks-") {
```
kalleep, line 56: Maybe we should handle both formats here for now? At least as long as we support writing them

claims/namespace.go
kalleep: Probably easier to just use [strings.CutPrefix](https://pkg.go.dev/strings#CutPrefix).
```go
if id, ok := strings.CutPrefix(ns, "stacks-1"); ok {
  stackId, err := strconv.Atoi(id)
  // ....
}
```

claims/namespace.go
charandas: Unintentional. Will revert. The block probably came from another repo where I was doing a similar check.

authz/namespace.go
```diff
@@ -26,15 +27,26 @@ var (
...
+
+const (
+	NamespaceAccessCheckerTypeCloud NamespaceAccessCheckerType = iota + 1
+	NamespaceAccessCheckerTypeOrg
+)
+
 type NamespaceAccessChecker interface {
-	CheckAccess(caller claims.AuthInfo, stackID int64) error
+	CheckAccess(caller claims.AuthInfo, namespace string) error
+	CheckAccessForIdentitfier(caller claims.AuthInfo, id int64) error
```
gamab, line 39: Personal preference:
```go
	CheckAccessForId(caller claims.AuthInfo, id int64) error
```
or
```go
	CheckAccessById(caller claims.AuthInfo, id int64) error
```

authz/namespace.go
```diff
@@ -79,15 +101,17 @@ func NewNamespaceAccessChecker(namespaceFmt claims.NamespaceFormatter, opts ...N
...
 		idClaims := caller.GetIdentity()
 		if idClaims == nil || idClaims.IsNil() {
 			if na.idTokenRequired {
 				return ErrorMissingIDToken
 			}
-		} else if idClaims.Namespace() != expectedNamespace {
+			// for else-if branch below,
+			// when id token claims are evaluated with an access token claims (with wildcard namespace) present
+			// but expectedNamespace is *, we skip the namespace equality check since it will always fail
+		} else if expectedNamespace != "*" && !checkEqualsNamespaceDisambiguous(expectedNamespace, idClaims.Namespace(), na.checkerType) {
```
gamab, line 114: If there is an id token, then AuthInfo represents a user. If `expectedNamespace == "*"` we should hence fail as I don't think a user can access all namespaces.

authz/namespace.go
```diff
@@ -79,15 +101,17 @@ func NewNamespaceAccessChecker(namespaceFmt claims.NamespaceFormatter, opts ...N
 	return na
 }
 
-func (na *NamespaceAccessCheckerImpl) CheckAccess(caller claims.AuthInfo, stackID int64) error {
-	expectedNamespace := na.namespaceFmt(stackID)
+func (na *NamespaceAccessCheckerImpl) CheckAccess(caller claims.AuthInfo, expectedNamespace string) error {
```
gamab, line 104: I think we can do something simpler to disambiguate the namespace:
```go
// disambiguousNamespace is a helper to temporarily navigate the issue with cloud namespace claims being ambiguous (stack vs stacks).
func disambiguousNamespace(namespace string) string {
	return strings.Replace(namespace, "stack-", "stacks-", 1)
}

func (na *NamespaceAccessCheckerImpl) CheckAccess(caller claims.AuthInfo, expectedNamespace string) error {
	expectedNamespace = strings.Replace(expectedNamespace, "stack-", "stacks-", 1)
	if na.idTokenEnabled {
		idClaims := caller.GetIdentity()
		if idClaims == nil || idClaims.IsNil() {
			if na.idTokenRequired {
				return ErrorMissingIDToken
			} else {
				namespace := disambiguousNamespace(idClaims.Namespace())
				if namespace != expectedNamespace {
					return ErrorIDTokenNamespaceMismatch
				}
			}
		}
	}
	if na.accessTokenEnabled {
		accessClaims := caller.GetAccess()
		if accessClaims == nil || accessClaims.IsNil() {
			return ErrorMissingAccessToken
		}
		namespace := disambiguousNamespace(accessClaims.Namespace())
		if namespace != "*" && namespace != expectedNamespace {
			return ErrorAccessTokenNamespaceMismatch
		}
	}
	return nil
}
```

charandas: I like it @gamab. Thanks for sharing, will adapt to using your snippet.

authz/namespace.go
gamab: @kalleep to confirm. But this does sound weird to me that we'd act on behalf of a user in the global namespace.

kalleep (approved):

PR #78, opened 2024-08-27
fix: resolves issues with broken discovery in Grafana

charandas:
When returning errors from `ParseNamespace`, we ran into broken auth when reaching `/apis` on an instance of Grafana.

This fixes it.

### DIFF
--- a/claims/namespace.go
+++ b/claims/namespace.go
@@ -57,15 +57,19 @@ func ParseNamespace(ns string) (NamespaceInfo, error) {
 		return info, nil
 	}
 
-	if strings.HasPrefix(ns, "org-") {
-		id, err := strconv.ParseInt(ns[4:], 10, 64)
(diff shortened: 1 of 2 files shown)


ryantxu (approved):

claims/namespace.go
```diff
@@ -57,15 +57,19 @@ func ParseNamespace(ns string) (NamespaceInfo, error) {
 		return info, nil
 	}
 
-	if strings.HasPrefix(ns, "org-") {
-		id, err := strconv.ParseInt(ns[4:], 10, 64)
-		if id < 1 {
+	if id, ok := strings.CutPrefix(ns, "org-"); ok {
```
kalleep, line 60: We did not touch this path in the other pr (an oversight). So this should have been broken even before this fix right?

PR #81, opened 2024-08-28
[authn] add Verifier unsafe mode

drclau:
This PR adds an unsafe implementation of the `Verifier` interface, `UnsafeVerifierBase`, which allows cracking the tokens without actually checking the signature.
This is meant to be used in Unified Storage for in-process mode.

### DIFF
--- a/authn/verifier.go
+++ b/authn/verifier.go
@@ -16,19 +16,41 @@ const (
 	TokenTypeAccess TokenType = "at+jwt"
 )
 
+type VerifierOption func(*VerifierBase[any])
+
+func WithKeyRetriever(keys KeyRetriever) VerifierOption {
+	return func(v *VerifierBase[any]) {
+		v.keys = keys
+	}
+}
+
+func WithUnsafeMode() VerifierOption {
+	return func(v *VerifierBase[any]) {
+		v.unsafeMode = true
+	}
+}
+
 type Verifier[T any] interface {
 	// Verify will parse and verify provided token, if `AllowedAudiences` was configured those will be validated as well.
 	Verify(ctx context.Context, token string) (*Claims[T], error)
 }
 
-func NewVerifier[T any](cfg VerifierConfig, typ TokenType, keys KeyRetriever) *VerifierBase[T] {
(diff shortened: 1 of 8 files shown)


gamab (approved):
LGTM! Best to change the `NewVerifier` signature with options sooner than later :+1:

authn/verifier.go
```diff
@@ -16,19 +16,41 @@ const (
...
+}
+
 type Verifier[T any] interface {
 	// Verify will parse and verify provided token, if `AllowedAudiences` was configured those will be validated as well.
 	Verify(ctx context.Context, token string) (*Claims[T], error)
 }
 
-func NewVerifier[T any](cfg VerifierConfig, typ TokenType, keys KeyRetriever) *VerifierBase[T] {
-	return &VerifierBase[T]{cfg, typ, keys}
+func NewVerifier[T any](cfg VerifierConfig, typ TokenType, opts ...VerifierOption) *VerifierBase[T] {
```
kalleep, line 38: Here we are mixing fuctional options with a config struct, we should stick with one way

kalleep (commented):
I think I would prefer to have this as an seperate implementation of Verifier interface instead of an option on `VerifierBase`, WDYT?

drclau:
> I think I would prefer to have this as an seperate implementation of Verifier interface instead of an option on `VerifierBase`, WDYT?

Sounds good to me. PTAL.

authn/verifier.go
```diff
@@ -42,6 +42,14 @@ func (v *VerifierBase[T]) Verify(ctx context.Context, token string) (*Claims[T],
...
 	}
 
+	claims := Claims[T]{
+		token: token, // hold on to the original token
+	}
+
+	if v.keys == nil {
+		return nil, errors.New("KeyRetriever is required for safe mode")
+	}
+
```
kalleep, line 52: I this change needed?

kalleep (approved):
Nice, much cleaner. And we are actually taking advantage of using our interfaces :+1:

gamab (approved):
Even better! Nice!

PR #83, opened 2024-08-30
[authn] add `NewUnsafeGrpcAuthenticator`

drclau:
Follow-up to https://github.com/grafana/authlib/pull/81

This PR adds `NewUnsafeGrpcAuthenticator` constructor, which uses of the `NewUnsafeIDTokenVerifier` and `NewUnsafeAccessTokenVerifier`.
This is meant to be used in Unified Storage for in-process mode.

### DIFF
--- a/authn/grpc_authenticator.go
+++ b/authn/grpc_authenticator.go
@@ -34,7 +34,7 @@ type GrpcAuthenticatorConfig struct {
 	IDTokenMetadataKey string
 
 	// KeyRetrieverConfig holds the configuration for the key retriever.
-	// Ignored if KeyRetrieverOption is provided.
+	// Ignored if KeyRetrieverOption is provided or when unsafe verifiers are used via NewUnsafeGrpcAuthenticator.
 	KeyRetrieverConfig KeyRetrieverConfig
 	// VerifierConfig holds the configuration for the token verifiers.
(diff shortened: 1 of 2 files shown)


drclau:
> Yes we can either do it like this or just make `NewGrpcAuthenticator` accept verifies. That we have to option to reuse one verifier in multiple places if we need.

@kalleep, I went for this approach because it is more explicit (i.e. you know what it means when you decide to use `NewUnsafeGrpcAuthenticator`), it's unlikely we'll need to support external verifiers, and the only existing use case for the unsafe verifiers is currently for unistor in-proc mode.
The alternative that I like is to pass the verifier as an option, and use a default if it's not set.
I actually don't have a strong preference, so if you think the latter is better suited long-term, I'll update the implementation. WDYT?

gamab (approved):
Works for me, clear and explicit. :+1: 

Maybe do both. ~~Add an option to specify the verifier and then in NewUnsafeGrpcAuthenticator call NewGrpcAuthenticator with an unsafe verifier. Wdyt?~~ We have two verifiers actually so that might not look pretty.

kalleep (approved):

drclau:
Thanks both, will merge as is, and we can iterate.

PR #90, opened 2024-09-10
GRPC/Authenticator: Add a simple local authenticator

ryantxu:
When using a local authenticator, we can just pull existing AuthInfo from the context.

### DIFF
--- a/authn/grpc_authenticator.go
+++ b/authn/grpc_authenticator.go
@@ -5,15 +5,19 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/grafana/authlib/claims"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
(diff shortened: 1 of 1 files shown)


authn/grpc_authenticator.go
```diff
@@ -5,23 +5,32 @@ import (
...
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
-
-	"github.com/grafana/authlib/claims"
 )
 
+// ContextAuthenticator ensures that an claims.AuthInfo exists in the context
```
ryantxu, line 17: replace https://github.com/grafana/grafana/blob/main/pkg/services/grpcserver/interceptors/auth.go#L20

ryantxu:
Looking more -- i think https://github.com/grafana/grafana/pull/93160 is likely a better option

PR #95, opened 2024-09-24
use new namespace format

kalleep:
The change to allow both `stack` and `stacks` prefix have been rolled out so we are safe to change this

Part of: https://github.com/grafana/identity-access-team/issues/843


drclau (approved):
🚀

gamab (approved):

Jguer (approved):

PR #98, opened 2024-09-26
Adjust AccessClient interface

kalleep:
I want us to get to a state where we have a client interface we can commit to.

As a minimum I think we need something like I have suggested in this pr. There are still some thoughts about some permissions we have that maybe don't fit this standard.

1. First I changed the name of `HasAccess` to `Check` to more closely map it to zanzana. This is an operation that will perform one check for a specific resource.
2. I added `List`. Under the hood this is probably the same as `Compile` without creating a function. I talked a bit with @alexanderzobnin and this is a valid strategy to use if we have few resources when doing filtering.

I added separate Request and Response structures. For list request we don't need to provide a name just the resource type and it is also useful to wrap the response in a structure if we need to add more to it.

There is also a separate function I think we need that I for now call "Capabilities". The idea here would be to query this for a specific resource and get back what actions can be performed on that resource but I have left it out for now because I don't know if it belongs on this interface or not

### DIFF
--- a/claims/access.go
+++ b/claims/access.go
@@ -2,6 +2,14 @@ package claims
 
 import "context"
 
+type Contextual struct {
+	// ~Kind eg dashboards
+	Resource string
+	// The specific resource
+	// In grafana, this was historically called "UID", but in k8s, it is the name
+	Name string
+}
+
 // CheckRequest describes the requested access.
 // This is designed bo to play nicely with the kubernetes authorization system:
 // https://github.com/kubernetes/kubernetes/blob/v1.30.3/staging/src/k8s.io/apiserver/pkg/authorization/authorizer/interfaces.go#L28
@@ -29,6 +37,10 @@ type CheckRequest struct {
 
 	// For non-resource requests, this will be the requested URL path
 	Path string
+
+	// Contextuals are additional resource + name that should be checked.
+	// E.g. for dashboards this can be the folder that a it belong to.
+	Contextuals []Contextual
 }
 
 type CheckResponse struct {
--- a/claims/access.go
+++ b/claims/access.go
@@ -2,10 +2,10 @@ package claims
 
 import "context"
 
-// AccessRequest describes the requested access.
+// CheckRequest describes the requested access.
 // This is designed bo to play nicely with the kubernetes authorization system:
 // https://github.com/kubernetes/kubernetes/blob/v1.30.3/staging/src/k8s.io/apiserver/pkg/authorization/authorizer/interfaces.go#L28
-type AccessRequest struct {
+type CheckRequest struct {
 	// The requested access verb.
 	// this includes get, list, watch, create, update, patch, delete, deletecollection, and proxy,
 	// or the lowercased HTTP verb associated with non-API requests (this includes get, put, post, patch, and delete)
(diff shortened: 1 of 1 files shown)


claims/access.go
```diff
@@ -31,15 +31,51 @@ type AccessRequest struct {
...
+	// Optional subresource
+	Subresource string
+
+	// For non-resource requests, this will be the requested URL path
+	Path string
+}
+
+type ListResponse struct {
+	// If all is set we can access all resources
+	All bool
```
alexanderzobnin, line 62: That's a good thing to have. It could be much cheaper check then.

kalleep: Yeah to only worry I have with this one is if this can be done trough openfga. I don't think so

kalleep:
I added `Contextuals` as part of `CheckRequest`. This would be used to pass e.g. a folder that a dashboard belong to. This would be required if we went with the approach where we don't store leaf nodes relation to its parent.

If we went with that approach how would we solve List / Compile? If we don't store e.g. parent for a dashboard then when doing a list request we would only get dashboards that we have direct relationship to.

The only thing I can think of would be to do two request, one to permissions directly to resource and one to list all permissions to folders and then checking that client side. E.g. `AccessChecker` would need to check both dashboard and its parent folder.

claims/access.go
ryantxu: I think it is still good to keep -- even if openfga could not (yet?) support it.  and how we model full access is still TBD

claims/access.go
```diff
@@ -29,17 +37,57 @@ type AccessRequest struct {
...
+	Namespace string
+
+	// Optional subresource
+	Subresource string
+
+	// For non-resource requests, this will be the requested URL path
+	Path string
+}
+
+type ListResponse struct {
```
ryantxu, line 72: I think keeping this as an interface is better than as a struct.  

Do we really want to commit to everything in one payload `Items []string`  with an interface, this could easily be backed by this struct response, or something that returns a bloomfilter and then pages?

claims/access.go
ryantxu: On top of that, if this requires contextuals, each list item will also need to inclue them.  so it would be more like:
```
type ItemChecker func(namespace string, name string, extra ...Contextual) bool
```

claims/access.go
```diff
@@ -2,10 +2,18 @@ package claims
 
 import "context"
 
-// AccessRequest describes the requested access.
+type Contextual struct {
+	// ~Kind eg dashboards
+	Resource string
```
ryantxu, line 7: Beyond folders, i'm not yet sure what we aim to represent.

for a k8s resource we would also need Group+Namespace (though we could enforce it is the same namespace as the original request)

claims/access.go
```diff
@@ -29,17 +37,57 @@ type AccessRequest struct {
...
+
+type CheckResponse struct {
+	Allowed bool
+}
+
+type ListRequest struct {
+	// The requested access verb.
+	// this includes get, list, watch, create, update, patch, delete, deletecollection, and proxy,
+	// or the lowercased HTTP verb associated with non-API requests (this includes get, put, post, patch, and delete)
+	Verb string
```
ryantxu, line 54: 🤔 maybe I am not understanding ListRequest and when it would be used.  

I assumed it was for the List verb

claims/access.go
kalleep: At least right now this is probably only going to be folders

claims/access.go
kalleep: I think it highly depends on how this is going to be used. @alexanderzobnin are exploring different strategies to do filtering. One of then would be to build a list and supply that as a filter for sql queries. This is a viable strategy if a user has few permissions but quickly falls apart

kalleep: We could also leave this out for now and only add it if / when we need it

claims/access.go
kalleep: My thinking was that this could be used to check for other permissions than just read access, but I removed it for now. Adding it back would be backwards compatible because you could always default to read

kalleep:
@ryantxu I am not sure what we are going to use `Group` and `Path` for in the context of checking access. Is the idea here that several apis could reuse names and we would use group as a way to make a distinction?

kalleep:
After discussing this with @ryantxu I will close this pr. Client used internally in services will probably not be this one but one that can solve contextual stuff.

I will experiment with an client internally in grafana instead and when we are happy with it we can probably move it out to authlib

PR #99, opened 2024-10-07
AuthZ: Multi-tenant client only requires one parent

gamab:
This PR refactors the Multi-Tenant AccessControl client to use a new `check` endpoint instead of the `read` endpoint. This will allow us not to require the entire list of parents of a given resource. The AuthZ Service (then Grafana) will handle the expansion of the folder tree.

**Changes**
- Migrated the client to use the `/check ` endpoint of the Authz service.
- Updated the `CheckRequest` struct to be more aligned with k8s: `namespace`, `resource`, `attribute`, `name`, and `parent` fields.

**Benefits:**
- The client will be simpler to use

**Next steps:**
- Update the documentation to reflect the changes to the RBAC system.
- Implement the server side of this change.


authz/multitenant_client.go
```diff
@@ -21,20 +22,21 @@ import (
...
 type CheckRequest struct {
-	Caller     claims.AuthInfo
-	StackID    int64
-	Action     string
-	Resource   *Resource
-	Contextual []Resource
+	Caller    claims.AuthInfo
+	Namespace string
+	Action    string
+	Object    string
```
gamab, line 38: Should I split this:
```suggestion
	// ~Kind eg dashboards
	Resource string
	// The specific resource
	// In grafana, this was historically called "UID", but in k8s, it is the name
	Name string
```

proto/v1/authz.proto
```diff
@@ -34,3 +47,16 @@ message ReadResponse {
   repeated Data data = 1;
   bool Found = 2;
 }
+
+message CheckRequest {
+  string namespace = 1;
+  string subject = 2;
+  string action = 3;
+  string object = 4;
```
gamab, line 55:
```suggestion
  string scope = 4;
```

authz/multitenant_client.go
```diff
@@ -21,20 +22,26 @@ import (
...
+	ErrReadPermission   = status.Errorf(codes.PermissionDenied, "read permission failed")
 )
 
 type CheckRequest struct {
-	Caller     claims.AuthInfo
-	StackID    int64
-	Action     string
-	Resource   *Resource
-	Contextual []Resource
+	Caller    claims.AuthInfo
```
gamab, line 35: Nit. Pass as a separate param

authz/multitenant_client.go
```diff
@@ -21,20 +22,26 @@ import (
...
+	Namespace string
+	Action    string
+	// ~Kind eg dashboards
+	Resource string
+	// The specific resource
```
gamab, line 40:
```suggestion
	// Attribute used to identify the resource in the legacy RBAC system.
	Attribute string
	// The specific resource
```

Jguer (approved):
LGTM, 

No need to reply but thinking points on my side:
- Will we need to keep a list of what's pareantable in the server side to make sure only resources that can have parents are checked this way.
- Validate that the parent can only be a folder without accessing the folder store ( the access would be denied anyway but how can we make sure the users of the platform don't think they can check annotations with parent dashboard)

authz/multitenant_client.go
```diff
@@ -21,24 +21,34 @@ import (
...
-	ErrMissingSubject = status.Errorf(codes.Unauthenticated, "missing subject")
-	ErrReadPermission = status.Errorf(codes.PermissionDenied, "read permission failed")
+	ErrMissingConfig           = errors.New("missing config")
+	ErrMissingRequestNamespace = status.Errorf(codes.InvalidArgument, "missing request namespace")
+	ErrInvalidRequestNamespace = status.Errorf(codes.InvalidArgument, "invalid request namespace")
+	ErrMissingRequestAttribute = status.Errorf(codes.InvalidArgument, "missing request attribute")
+	ErrMissingRequestResource  = status.Errorf(codes.InvalidArgument, "missing request resource")
+	ErrMissingRequestAction    = status.Errorf(codes.InvalidArgument, "missing request action")
+	ErrMissingCaller           = status.Errorf(codes.Unauthenticated, "missing caller")
+	ErrMissingSubject          = status.Errorf(codes.Unauthenticated, "missing subject")
```
kalleep, line 31: Any reason we are returning grpc errors fromt he client?

kalleep (commented):
Looks good, left a couple of nits / questions

authz/multitenant_client.go
```diff
@@ -21,24 +21,34 @@ import (
...
 type CheckRequest struct {
-	Caller     claims.AuthInfo
-	StackID    int64
-	Action     string
-	Resource   *Resource
-	Contextual []Resource
+	// The namespace in which the request is made (e.g. "stacks-12")
+	Namespace string
+	// The requested action (e.g. "dashboards:read")
+	Action string
```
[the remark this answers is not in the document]
kalleep: @gamab and me talked a bit about it yesterday. Initially I said we could probably skip it but I think it makes sense to do if  you both agree

authz/multitenant_client.go
gamab: No valid enough reason :D
Just that it embedded a status code that could be bubbled up to the end user. A bit like our Grafana errors.
Should I do a new error type or just ignore the status code altogether?

authz/multitenant_client.go
```diff
@@ -21,24 +21,34 @@ import (
...
+	// Attribute used to identify the resource in the legacy RBAC system (e.g. uid).
+	Attribute string
+	// The specific resource
+	// In grafana, this was historically called "UID", but in k8s, it is the name
+	Name string
+	// The Name of the parent folder of the resource
+	Parent string
 }
 
 type MultiTenantClient interface {
```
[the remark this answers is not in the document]
gamab: As opposed to the `EnforcementClient` that the `authz` library also exposes, this client can work with multiple organizations on-prem or multiple stacks in cloud.
I have no strong attachement to this name, but I think it's more explicit than `AccessClient`.
Can you please suggest alternatives?

authz/multitenant_client.go
kalleep: I think regular errors with `errors.New` are fine

authz/multitenant_client.go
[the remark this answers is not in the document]
kalleep: Yes sure that is true but do we plan to keep `EnforcementClient` around? I think this should be the only client at least as an interface.

authz/multitenant_client.go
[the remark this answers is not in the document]
gamab: `list / get` => `read`
`post`  => `create`
`put` => `write`
`delete` => `delete`

kalleep (approved):

PR #100, opened 2024-10-10
Document: Client interface

gamab:


authz/multitenant_client.go
```diff
@@ -45,7 +45,19 @@ type CheckRequest struct {
...
+	//
+	// Caller represents the authentication information of the entity
+	// initiating the request, which can be a service or a user.
+	//
+	// CheckRequest contains the details of the request, including the namespace, action, resource, parent.
+	//
+	// The method returns true if the Caller is authorized, false otherwise.
+	// An error is returned if the authorization check cannot be completed,
+	// for example, due to an unreachable authorization service.
 	Check(ctx context.Context, Caller claims.AuthInfo, req *CheckRequest) (bool, error)
```
kalleep, line 61: Is it to late to change Check to return something like
```go
type CheckResponse struct {
   Allowed bool
 }
```

Then if we ever need to return more information we don't have to break the interface but can just add it to the return struct

kalleep (approved):
Nice really good documentation. Added an unrelated comment that should have been on previous pr. If we don't want to do it I think its fine either way

PR #102, opened 2024-10-10
Remove unecessary tracing

gamab:


kalleep (commented):
There is still usage in `wrapContext`, the property and option to set tracer. We should clean those up as well

drclau:
> There is still usage in `wrapContext`, the property and option to set tracer. We should clean those up as well

@gamab, is the intention here to remove tracing support altogether, or just to tidy up things?

drclau (approved):
🚀

PR #105, opened 2024-10-16
AuthZ: Implement access client

gamab:
This PR is a proposal to replace the Client interface with the AccessClient altogether.

The most important change here is the use of `Group` (e.g. `dashboards.grafana.app`), `Resource` (e.g. `dashboads`), `Verb` (e.g. `list`), in place of the RBAC action.
This change implies:
* some mapping will be done on the server side to enforce access appropriately.
* access tokens will have to be defined with both the kubernetes fully qualified verb (e.g: `dashboards.grafana.app/dashboards:list`) **and** the RBAC action **if they need to query Grafana**. Until we decide if we want to solve the mapping client side as well.


authz/client.go
```diff
@@ -219,92 +201,157 @@ func (r *CheckRequest) Validate() error {
 	return nil
 }
 
-func (c *LegacyClientImpl) Check(ctx context.Context, caller claims.AuthInfo, req *CheckRequest) (CheckResponse, error) {
-	ctx, span := c.tracer.Start(ctx, "LegacyClientImpl.Check")
+// Compile implements claims.AccessClient.
+func (c *ClientImpl) Compile(ctx context.Context, id claims.AuthInfo, req claims.AccessRequest) (claims.AccessChecker, error) {
```
kalleep, line 205: It is possible to implement this right? I guess you have not gotten around too it

kalleep (commented):
Nice, I still think it makes sense to do some of the changes from https://github.com/grafana/authlib/pull/98.

Mostly around naming but not structures.
1. Change `HasAccess` to `Check` and AccessRequest to `CheckRequest` and return CheckResponse instead of just bool.
2.  Change request on Compile from AccessRequest to ListRequest that has everything from CheckRequest except `Name` because name does not make sense for that function. Internally Compile will do a list request and then create the Checker function. Then if we later want to ever expose a List function we just use the same structure

authz/client.go
kalleep, line 205: This is possbile to implement right?

gamab: `impossible`?

kalleep: Without expansion is possible right?

kalleep: If not we should probably split up `AccessClient` into two interfaces and only implement one for now. I don't think we chould expose a function that can only panic

gamab: Ok, wanted to make sure your question was correct.

We cannot do folder expansion in an efficient way I think, yes. So if I implement something while we wait for the dual writes, it'd be a compile without expansion.

> If not we should probably split up AccessClient into two interfaces and only implement one for now. 

Sounds good to me :+1:

authz/client.go
```diff
@@ -194,62 +159,96 @@ func NewLegacyClient(cfg *ClientConfig, opts ...LegacyClientOption) (*LegacyClie
...
+
+	checkReq := &authzv1.CheckRequest{
+		Subject:     id.GetIdentity().Subject(),
+		Verb:        req.Verb,
+		Group:       req.Group,
+		Resource:    req.Resource,
+		Namespace:   req.Namespace,
+		Name:        req.Name,
+		Subresource: req.Subresource,
+		Path:        req.Path,
```
gamab, line 202: We won't support those in the beginning. Not yet even sure how Path should be managed. But adding them not to have to modify the proto file later if they are needed.

gamab:
> Change HasAccess to Check and AccessRequest to CheckRequest and return CheckResponse instead of just bool.
I expected you to say so :smile:
Change done.

> Change request on Compile from AccessRequest to ListRequest that has everything from CheckRequest except Name because name does not make sense for that function. Internally Compile will do a list request and then create the Checker function. Then if we later want to ever expose a List function we just use the same structure

Done.

claims/access.go
```diff
@@ -29,17 +29,50 @@ type AccessRequest struct {
...
+type AccessChecker interface {
+	// Check checks whether the user can perform the given action for all requests
+	Check(ctx context.Context, id AuthInfo, req CheckRequest) (CheckResponse, error)
+}
+
+type ListRequest struct {
+	// The requested access verb.
+	// this includes get, list, watch, create, update, patch, delete, deletecollection, and proxy,
+	// or the lowercased HTTP verb associated with non-API requests (this includes get, put, post, patch, and delete)
+	Verb string
```
ryantxu, line 51:
```suggestion
```
For a list...  there is not really a verb different than "list" -- and we expect to be able to check if we can read/view each item.

The same would be used for search queries

claims/access.go
```diff
@@ -29,17 +29,50 @@ type AccessRequest struct {
...
+	Namespace string
+
+	// Optional subresource
+	Subresource string
+}
+
+// TODO: Should the namespace be specified in the request instead.
+// I don't think we'll be able to Compile over multiple namespaces.
+// Checks access while iterating within a resource
+type ItemChecker func(namespace string, name string) bool
```
ryantxu, line 69:
```suggestion
type ItemChecker func(namespace, name, folder string) bool
```

each item will need a folder check 😢

claims/access.go
```diff
@@ -29,17 +29,50 @@ type AccessRequest struct {
...
 type AccessClient interface {
-	// HasAccess checks whether the user can perform the given action for all requests
-	HasAccess(ctx context.Context, id AuthInfo, req AccessRequest) (bool, error)
+	AccessChecker
 
 	// Compile generates a function to check whether the id has access to items matching a request
 	// This is particularly useful when you want to verify access to a list of resources.
 	// Returns nil if there is no access to any matching items
-	Compile(ctx context.Context, id AuthInfo, req AccessRequest) (AccessChecker, error)
+	Compile(ctx context.Context, id AuthInfo, req ListRequest) (ItemChecker, error)
```
ryantxu, line 77: looking more... maybe we should return a struct/interface that could also supports fetching all IDs (paged?)

claims/access.go
```diff
@@ -29,17 +29,50 @@ type AccessRequest struct {
...
+	// Folder is the parent folder of the requested resource
+	Folder string
+}
+
+type CheckResponse struct {
+	// Allowed is true if the request is allowed, false otherwise.
+	Allowed bool
 }
 
+type AccessChecker interface {
```
ryantxu, line 42: now that we know this client is only aims to be used by trusted api-servers, and *not* a generic top-level client for everything.   Should we move it back to the authz package?

ryantxu:
This is looking great!

claims/access.go
kalleep: Yes :rocket:

claims/access.go
kalleep: We could also just expose List as a method. It is essentially the same as `Compile` without building the `ItemChecker`

gamab: Let's maybe tackle this discussion separately.
I think this will need a bit of work to get right, specially with folder. Do we need to make two call, one for the resource rules, one for the folder rules? Do we need to handle pagination? (I personally think we should hide that complexity as much as possible).

authz/client.go
```diff
@@ -78,11 +111,13 @@ type ClientConfig struct {
 	accessTokenAuthEnabled bool
 }
 
-var _ Client = (*LegacyClientImpl)(nil)
+// ClientImpl will implement the claims.AccessClient interface
+// Once we are able to deal with folder permissions expansion.
+var _ AccessChecker = (*ClientImpl)(nil)
 
-type LegacyClientOption func(*LegacyClientImpl)
+type LegacyClientOption func(*ClientImpl)
```
kalleep, line 118: We should rename the options

authz/client.go
```diff
@@ -194,62 +229,96 @@ func NewLegacyClient(cfg *ClientConfig, opts ...LegacyClientOption) (*LegacyClie
...
+		return false, err
+	}
+
+	// Cache the result
+	err = c.cacheCheck(ctx, key, resp.Allowed)
+
+	return resp.Allowed, err
+}
+
+// HasAccess implements claims.AccessClient.
```
kalleep, line 291:
```suggestion

```

authz/client.go
gamab: I renamed it `AuthZClientOption` but I'd prefer `ClientOption`. But that'd break plugins updating the lib.

kalleep (approved):
All good from my side

PR #109, opened 2024-11-04
AuthZ: Set expiration on cache client

gamab:
Set cache entries expiration to 5 minutes (currently they don't expire) and clean up interval to 1 minute.

### DIFF
--- a/authz/client.go
+++ b/authz/client.go
@@ -6,6 +6,7 @@ import (
 	"encoding/gob"
 	"errors"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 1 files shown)


eleijonmarck (approved):
⚡ 
nit: what is the cleanup process? running it every minute, could be CPU intensive?

gamab:
>  what is the cleanup process? running it every minute, could be CPU intensive?

It's cleaning up expired entries to lower memory usage. I don't know how CPU intensive this is. Maybe I should use the 10 minutes default Grafana uses.

PR #111, opened 2024-11-07
Tracing: Don't propagate span in context

gamab:
Avoid propagating the span in context. It's otherwise considered as parent of the following spans (even if it's ended).

image


authn/grpc_authenticator.go
```diff
@@ -153,19 +153,19 @@ func NewUnsafeGrpcAuthenticator(cfg *GrpcAuthenticatorConfig, opts ...GrpcAuthen
...
-	md, ok := metadata.FromIncomingContext(ctx)
+	md, ok := metadata.FromIncomingContext(spanCtx)
 	if !ok {
 		span.RecordError(ErrorMissingMetadata)
 		return nil, ErrorMissingMetadata
 	}
 
 	if ga.cfg.accessTokenAuthEnabled {
-		atClaims, err := ga.authenticateService(ctx, md)
+		atClaims, err := ga.authenticateService(spanCtx, md)
```
gamab, line 168: Not actually useful for now, but might be one day.

kalleep (approved):

PR #113, opened 2024-11-21
Fix: refactor auth info

kalleep:
Sorry for this big pr but It was necessary in order to make the interface easier to work with.

When performing authentication using access and id tokens we have two cases.

1. Authenticate as a service using only access token
2. Authenticate on behalf of the user using both access token and id token

Before the interface was composed from IdentityClaims and AccessClaims. We had to have `IsNil` in order to handle cases where e.g. authentication was done with only access tokens. 

I decided to flatten out the interface and add common functionality for them.

When authenticating only using a access token every field such as `GetUID`, `GetNamespace` etc will be extracted from that access token. When authenticating on behalf of a user we instead extract these fields from the id token because the acccess token act as that identity.

`GetTokenPermissions` and `GetTokenDelegatedPermissions` are alwas extracted from access token because id token does not have that information.

If you have any other suggestions I am open to them :). I will take care of updating relevant places within grafana and adjust our implementations of this interface there


claims/token.go
```diff
@@ -33,134 +42,33 @@ type AuthInfo interface {
...
-	// accidentally assigned to a different data object; if the application
-	// uses multiple issuers, collisions MUST be prevented among values
-	// produced by different issuers as well.  The "jti" claim can be used
-	// to prevent the JWT from being replayed.  The "jti" value is a case-
-	// sensitive string.  Use of this claim is OPTIONAL.
-	// [RFC 7519 §4.1.7]: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7
-	JTI() string
-}
+	// GetTokenPermissions returns Grafana permissions that authenticated access token can perform.
+	GetTokenPermissions() []string
```
kalleep, line 54: I can't use `GetPermissions` because it would collide with https://github.com/grafana/grafana/blob/main/pkg/apimachinery/identity/requester.go#L44

charandas: Will this be a useful function in Grafana backend when its just the user as themself who has authenticated using a browser cookie? I guess not?

Also, the fact that it doesn't say which token it corresponds to, might make it confusing for someone not familiar with the internals.

So I understand the use-case for service auth (both as itself and OBO user), in that this is a useful function but just curious what will communicate to the developer down the road that this `func` may or may not apply in their auth use-case, given that it will be leaked to the `requester` (since `AuthInfo` is directly mingled into the `requester` interface).

charandas: I recognize that it is not much different currently in that, requester contains both AccessClaims and IdentityClaims in its interface.

```go
type Requester interface {
	user.Info
	claims.AuthInfo // which is just a struct with both of those claims interfaces
	// ... rest of the interface omitted ...
}
```

charandas:
Also, it does look like `ext_jwt` implementation takes care of some of these same details [here](https://github.com/grafana/grafana/blob/main/pkg/services/authn/clients/ext_jwt.go#L186-L201), is it that one of these places will become the source of truth after this change, or will that implementation continue to construct an `authn.Identity` with replicated logic?

I see that `authn.AuthInfo` here is a struct, which implements the `claims.AuthInfo` interface here, and this struct is being newly added. Potentially, could be nice if Grafana started using this one struct instead of all the different identity implementations (`authn.Identity`, `user.Identity`) it has. Is that one of the reasons of introducing this struct?

The [AuthID](https://github.com/grafana/grafana/blob/main/pkg/services/authn/clients/ext_jwt.go#L194C1-L194C48) property on the Identity is an interesting one, it could let a developer differentiate an identity of the user that came from service acting OOB user versus a user as themself given that `AuthID` being different from `ID` would only happen in the former case. Are your changes here concerned with providing an AuthID getter so `authn.Identity`? It looks like `GetSubject` is that. `GetIdentifier()` being different from `GetSubject()` seems like it would help discern the OOB situation, and `GetIdentifier()` being the same as `GetSubject()` the user as themself.

claims/token.go
kalleep: For the first question, no it won't. This this function + `GetTokenDelegatedPermissions` are only useful when access tokens are used with or without an id token.

Everything else can be used in normal authentication flow. `Request` interface is internal to single tenant grafana. That interface cannot be used in apps / external services because it contains information we don't don't propagate down. Apps should only used AuthInfo. 

There are some things we can remove from `Requester` with this refactor because both this interface and that one have functions that do the same but with different names.

claims/token.go
```diff
@@ -1,24 +1,33 @@
...
 	// if the user is removed from the system and another user is added with
 	// the same name.
 	// This will be in the form: <IdentityType>:<Identifier>
 	GetUID() string
 
-	// GetGroups returns the names of the groups the user is a member of
-	// In grafana, this will only be populated with standard k8s names
+	// GetIdentifier returns only the Identitfier part.
+	// For some identity types this can be empty e.g. Anonymous.
+	GetIdentifier() string
```
Jguer, line 20: Will this be the ID or UID?

From the description of GetSubject I get the impression it's ID

authz/namespace.go
```diff
@@ -19,17 +19,12 @@ const (
...
-	ErrorMissingAccessToken           = status.Errorf(codes.Unauthenticated, "unauthenticated: missing access token")
-	ErrorIDTokenNamespaceMismatch     = status.Errorf(codes.PermissionDenied, "unauthorized: id token namespace does not match expected namespace")
-	ErrorAccessTokenNamespaceMismatch = status.Errorf(codes.PermissionDenied, "unauthorized: access token namespace does not match expected namespace")
+	ErrorMissingMetadata = status.Errorf(codes.Unauthenticated, "unauthenticated: missing metadata")
+	ErrNamespaceMismatch = status.Errorf(codes.PermissionDenied, "unauthorized: namespace does not match expected namespace")
 )
 
 type NamespaceAccessChecker interface {
 	CheckAccess(ctx context.Context, caller claims.AuthInfo, namespace string) error
-	CheckAccessByID(ctx context.Context, caller claims.AuthInfo, id int64) error
```
gamab, line 32: Thanks :+1:

kalleep:
> I see that authn.AuthInfo here is a struct, which implements the claims.AuthInfo interface here, and this struct is being newly added

That one existed before this pr. Its is used by the GRPC authenticator so I have to keep that implementation around. That implementation will / is used by unified storage.

Ideally we would use only this interface inside single tenant grafana as well but it is quite a lot of work to achieve that. But this pr tries to at least align these two interface more so its easier to start slim the other one down. Before this pr AuthInfo had two methods, GetAccess and GetIdentity and those makes even less sense in `Requester` because they are more tied to access token and id token claims but most auth methods in ST grafana don't use them.

 >The [AuthID](https://github.com/grafana/grafana/blob/main/pkg/services/authn/clients/ext_jwt.go#L194C1-L194C48) property on the Identity is an interesting one
 
AuthID is a bit different. Its is an external id for a user when performing authentication. E.g. when signing in with OAuth it is subject provided by the idp used. For access token the external id is the same as subject

claims/token.go
kalleep: Ideally we would remove `GetSubject` but right now we can't. Authz service can only work with internal id:s of grafana entities

claims/token.go
gamab: > I can't use GetPermissions because it would collide with https://github.com/grafana/grafana/blob/main/pkg/apimachinery/identity/requester.go#L44

:cry:

authz/namespace.go
```diff
@@ -100,43 +61,13 @@ func (na *NamespaceAccessCheckerImpl) CheckAccess(ctx context.Context, caller cl
...
-			return ErrorMissingAccessToken
-		}
-		// for if branch below,
-		// when access token claims with a wildcard namespace are passed in, we skip the namespace equality check
-		// it **will fail** when checking on resources in specific namespaces, which we don't want
-		if !claims.NamespaceMatches(accessClaims, expectedNamespace) {
-			span.RecordError(ErrorAccessTokenNamespaceMismatch)
-			return ErrorAccessTokenNamespaceMismatch
-		}
+	if expectedNamespace != "*" && !claims.NamespaceMatches(caller.GetNamespace(), expectedNamespace) {
```
gamab, line 64: This is problematic to me. An access token without `*` trying to access `*` should be rejected.

gamab (dismissed):
Great job! The interface changes make things easier from a caller perspective! :tada: 

~Currently for `in-proc grpc` on-prem, we consider that having no access-token is ok. But some of the changes made to the `authz` section rely on an access token being set which would be breaking.~

~We can either change authentication to generate an in-proc service token with all permissions. Or we wait for AT to be available on-prem but need to change the authz code to allow 'no-access token'.~

authn/auth_info.go
```diff
@@ -0,0 +1,125 @@
...
+		return a.id.Rest.Identifier
+	}
+	return strings.TrimPrefix(a.at.Subject, string(claims.TypeAccessPolicy)+":")
+}
+
+func (a *AuthInfo) GetIdentityType() claims.IdentityType {
+	if a.id != nil {
+		return a.id.Rest.Type
+	}
+	return claims.TypeAccessPolicy
```
gamab, line 54: Ah I missed this one

authz/client.go
```diff
@@ -379,27 +367,18 @@ func (c *ClientImpl) Check(ctx context.Context, id claims.AuthInfo, req CheckReq
 }
 
 func (c *ClientImpl) validateCaller(caller claims.AuthInfo) error {
-	accessClaims := caller.GetAccess()
-	if c.authCfg.accessTokenAuthEnabled && (accessClaims == nil || accessClaims.IsNil()) {
+	if caller.GetSubject() == "" {
```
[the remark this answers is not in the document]
kalleep: Fixed in [eaa004e](https://github.com/grafana/authlib/pull/113/commits/eaa004e67344bc0a2bb43820da4af96cc39ee157)

authz/client.go
```diff
@@ -379,27 +367,18 @@ func (c *ClientImpl) Check(ctx context.Context, id claims.AuthInfo, req CheckReq
...
-	// Check both AccessToken and IDToken (if present) for namespace match
-	accessClaims := caller.GetAccess()
-	accessTokenMatch := !c.authCfg.accessTokenAuthEnabled ||
-		(accessClaims != nil && !accessClaims.IsNil() && claims.NamespaceMatches(accessClaims, expectedNamespace))
-
-	idClaims := caller.GetIdentity()
-	idTokenMatch := idClaims == nil || idClaims.IsNil() || claims.NamespaceMatches(idClaims, expectedNamespace)
+	if !c.authCfg.accessTokenAuthEnabled && claims.IsIdentityType(caller.GetIdentityType(), claims.TypeAccessPolicy) {
+		return true
+	}
```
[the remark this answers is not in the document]
kalleep: So with the implementation of AuthInfo we have in authn package this would work because the default value is claims.TypeAccessPolicy

authz/namespace.go
kalleep: TBH we should probably remove this whole thing. The only thing it does is calling `claims.NamespaceMatches` and only used in one place. If they need to handle it differently we should handle it there I guess?

kalleep: yep it is, I remove that one. Will this be a problem for you @charandas

gamab (approved):
LGTM!

claims/token.go
```diff
@@ -1,24 +1,29 @@
...
-	// GetUID returns a unique value for a particular user that will change
+	// GetUID returns a unique value for a particular identity that will change
 	// if the user is removed from the system and another user is added with
 	// the same name.
 	// This will be in the form: <IdentityType>:<Identifier>
 	GetUID() string
 
-	// GetGroups returns the names of the groups the user is a member of
-	// In grafana, this will only be populated with standard k8s names
+	// GetIdentifier returns only the Identitfier part.
```
gamab, line 14:
```suggestion
	// GetIdentifier returns only the Identifier part.
```

authn/auth_info.go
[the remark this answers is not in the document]
kalleep: So if we have the need we can add it later. I just want to have what is required for now and we can expand / fix it later if that is okey?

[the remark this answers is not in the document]
gamab: Sounds good to me

PR #114, opened 2024-11-22
authn/claims: Fix segfaults in `AuthInfo.GetName()` and `AuthInfo.GetUID()`

onprem:
These methods currently segfault in case `IdentityClaims` is `nil`. I have added a nil check to avoid the segfaults.

I ran into this when using Unified Storage with the Cloud API Server which only passes an Access Token. The `GetUID()` method is used for creating deletion markers in US.

### DIFF
--- a/authn/claims.go
+++ b/authn/claims.go
@@ -45,12 +45,18 @@ func (c *AuthInfo) GetGroups() []string {
 
 // Name implements claims.AuthInfo.
 func (c *AuthInfo) GetName() string {
-	return c.IdentityClaims.claims.Rest.getK8sName()
+	if c.IdentityClaims != nil && !c.IdentityClaims.IsNil() {
+		return c.IdentityClaims.claims.Rest.getK8sName()
+	}
+	return ""
 }
 
 // UID implements claims.AuthInfo.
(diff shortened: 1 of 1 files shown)


authn/claims.go
```diff
@@ -45,12 +45,18 @@ func (c *AuthInfo) GetGroups() []string {
...
+	if c.IdentityClaims != nil && !c.IdentityClaims.IsNil() {
+		return c.IdentityClaims.claims.Rest.getK8sName()
+	}
+	return ""
 }
 
 // UID implements claims.AuthInfo.
 func (c *AuthInfo) GetUID() string {
-	return c.IdentityClaims.claims.Rest.asTypedUID()
+	if c.IdentityClaims != nil && !c.IdentityClaims.IsNil() {
```
onprem, line 56: Maybe we can extend the `AccessTokenClaims` [verifier_access_token.go#L7-L17]( https://github.com/grafana/authlib/blob/main/authn/verifier_access_token.go#L7-L17) to include Cloud Access Policy `ID`, which can be used for `GetUID()` here in case `IdentityClaims` is `nil`. Similarly using the CAP `Name` for `AuthInfo.GetName()`.

How does this sounds?

PR #119, opened 2024-12-11
AuthN: Add "Authenticator" capable of performing authentication on tokens

kalleep:
We discussed it a bit here https://raintank-corp.slack.com/archives/C06PXDT9Y85/p1733216056937919?thread_ts=1733166073.080389&cid=C06PXDT9Y85

The Idea here is to have a "Authenticator" interface that will authenticate tokens. I made it protocol agnostic by using `TokenProvider` and implemented it for GRPC and HTTP.

If we want to implement something like this we can replace most of the code in the grpc authenticator using this and the code here: https://github.com/grafana/grafana/blob/main/pkg/services/authn/clients/ext_jwt.go#L73

It is also something we can reuse in other places.

### DIFF
--- a/authn/authenticator.go
+++ b/authn/authenticator.go
@@ -60,7 +60,7 @@ type GRPCTokenProvider struct {
 }
 
 func (p GRPCTokenProvider) AccessToken(_ context.Context) (string, bool) {
-	const key = "X-Access-Token"
+	const key = DefaultAccessTokenMetadataKey
 	values := p.md.Get(key)
 	if len(values) == 0 {
 		return "", false


kalleep (commented):
Marking this a draft and left out documentation / tests. Don't want to invest to much time if we don't think we need it.

authn/authenticator.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to verify access token: %w", err)
+	}
+
+	return NewAccessTokenAuthInfo(*claims), nil
+}
+
+var _ Authenticator = (*IDTokenAuthenticator)(nil)
+
+func NewIDTokenAutenticator(id *IDTokenVerifier) *IDTokenAuthenticator {
```
kalleep, line 143: This one should not really exist but right now we can configure grpc authenticator to only authenticate id tokens...

authn/authenticator.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil, errors.New("unauthenticated")
+	}
+
+	atClaims, err := a.at.Verify(ctx, atToken)
+	if err != nil {
+		return nil, fmt.Errorf("failed to verify access token: %w", err)
+	}
+
+	idToken, ok := provider.IDToken(ctx)
+	if !ok {
```
charandas, line 100: this reads slightly weird even though I know what you mean here, "!ok being an ok case", of only access token being present.

kalleep: Yeah I should leave some comments to explain everything. I just need to figure out if this is something we need

authn/authenticator.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// DefaultAuthenticator will try to authenticate using both access token and id token.
+// Authnetication can be done with only a access token or with both access token and id token.
+type DefaultAuthenticator struct {
+	at *AccessTokenVerifier
+	id *IDTokenVerifier
+}
+
+func (a *DefaultAuthenticator) Authenticate(ctx context.Context, provider TokenProvider) (claims.AuthInfo, error) {
```
gamab, line 102: We're creating a new `<Protocol>TokenProvider` on each request, but the provider code itself doesn't change. I wonder if there is a way to optimize this?

gamab (approved):
LGTM.

authn/authenticator.go
```diff
@@ -0,0 +1,179 @@
...
+	if len(values) == 0 {
+		return "", false
+	}
+
+	token := values[0]
+	return token, len(token) > 0
+}
+
+func (p GRPCTokenProvider) IDToken(_ context.Context) (string, bool) {
+	const key = "X-Grafana-Id"
```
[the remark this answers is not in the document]
kalleep: Oh so you used something different for grpc then we do for forwarding to e.g. plugins etc?

authn/authenticator.go
```diff
@@ -0,0 +1,179 @@
...
+		return "", false
+	}
+
+	token := values[0]
+	return token, len(token) > 0
+}
+
+// Authenticator is used to authenticate request using the provided TokenProvider.
+type Authenticator interface {
+	Authenticate(ctx context.Context, provider TokenProvider) (claims.AuthInfo, error)
```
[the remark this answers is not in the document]
kalleep: I don't like that idea. IIUC you would then need e.g.
NewDefaultGRPCAuthenticator
NewDefaultHTTPAuthenticator

And have custom implementations for each one. I think this should be handled based on where we use them

authn/authenticator.go
kalleep: Not much to "optimize" you just wrap what you have in a structure and use that

authn/authenticator.go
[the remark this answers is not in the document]
gamab: Yep seems so

PR #118, opened 2024-12-11
AuthZ: Add support for wildcards

gamab:

PR #121, opened 2024-12-11
AuthZ: Add support for wildcards - Alternative

gamab:
Twice faster and more memory-efficient than #118 

Also more restricted in terms of pattern matching. 

```
BenchmarkHasPermissionInToken-12     	 3745948	       363.3 ns/op	     128 B/op	       5 allocs/op
BenchmarkHasPermissionInToken2-12    	 1542550	       732.3 ns/op	     208 B/op	       8 allocs/op
```

```
BenchmarkHasPermissionInToken-12              	 3905312	       324.0 ns/op	     128 B/op	       5 allocs/op
BenchmarkHasPermissionInToken_Wildcards-12    	 4503369	       368.4 ns/op	     128 B/op	       4 allocs/op
```

<details>

```go

func BenchmarkHasPermissionInToken(b *testing.B) {
	tokenPermissions := []string{
		"dashboard.grafana.app/dashboards:list",
		"dashboard.grafana.app/dashboards:get",
		"dashboard.grafana.app/dashboards:create",
		"dashboard.grafana.app/dashboards:delete",
		"dashboard.grafana.app/dashboards:update",
		"dashboard.grafana.app/folders:list",
		"dashboard.grafana.app/folders:get",
		"dashboard.grafana.app/folders:create",
		"dashboard.grafana.app/folders:delete",
		"dashboard.grafana.app/folders:update",
	}

	group := "dashboard.grafana.app"
	resource := "dashboards"
	verb := "get"
	name := "dashUID"

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if !hasPermissionInToken(tokenPermissions, group, resource, verb, name) {
			b.Fail()
		}
	}
}

func BenchmarkHasPermissionInToken_Wildcards(b *testing.B) {
	tokenPermissions := []string{
		"*.grafana.app/*:get",
	}

	group := "dashboard.grafana.app"
	resource := "dashboards"
	verb := "get"
	name := "dashUID"

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if !hasPermissionInToken(tokenPermissions, group, resource, verb, name) {
			b.Fail()
		}
	}
}
```

</details>


kalleep:
As mentioned earlier I am not sure I like the idea of introducing wildcard, I think prefix matching is way easier to reason about. But I get the case st grafana to unified storage have. E.g. allow `*.grafana.app`.

For the current implementation I found a case that surprisingly failed `pattern = *x input = xx` but `pattern=*x input=x` works. I would expect the first case to pass because it says match anything but end with `x`. This is failing for leading input that matches the pattern so `pattern = *testing input = testingtesting` also failes.

authz/client.go
```diff
@@ -292,6 +293,79 @@ func (c *ClientImpl) check(ctx context.Context, id claims.AuthInfo, req *CheckRe
...
+	// empty pattern only matches empty input
+	if len(pattern) == 0 {
+		return len(input) == 0
+	}
+
+	if pattern == "*" {
+		return true
+	}
+
+	patternParts := strings.FieldsFunc(pattern, func(r rune) bool { return r == '*' })
```
gamab, line 308: Todo Can be empty

IevaVasiljeva (approved):
LGTM! I'm fine with either this or the prefix matching implementation, but don't see any reasons against matching this (as it provides a bit more flexibility).

Nit: maybe add some more wildcard test cases that should not match (eg, wildcard group and name match, but resource doesn't match).

kalleep:
So wildcard solution would be much more flexible, you can express access in ways you can't with prefix matching. But it is also more complicated and probably less performat (would need to verify that). 

Matching on prefix would emulate how we can express things in zanzana so it would fit that model better

gamab:
> Matching on prefix would emulate how we can express things in zanzana so it would fit that model better

If for some reason we end up removing permissions from tokens and enforcing them with Zanzana then wildcards are potentially going to be hard to migrate. So dropping this one for #124. We can always re-open this :)

PR #124, opened 2024-12-17
AuthZ: Implement prefix matching

gamab:
```
BenchmarkHasPermissionInToken-12              	 1999384	       611.6 ns/op	     240 B/op	       9 allocs/op
BenchmarkHasPermissionInToken_Prefixs-12      	 2163127	       574.4 ns/op	     192 B/op	       8 allocs/op
```

<details>

```go
func BenchmarkHasPermissionInToken(b *testing.B) {
	tokenPermissions := []string{
		"dashboard.grafana.app/dashboards:list",
		"dashboard.grafana.app/dashboards:get",
		"dashboard.grafana.app/dashboards:create",
		"dashboard.grafana.app/dashboards:delete",
		"dashboard.grafana.app/dashboards:update",
		"dashboard.grafana.app/folders:list",
		"dashboard.grafana.app/folders:get",
		"dashboard.grafana.app/folders:create",
		"dashboard.grafana.app/folders:delete",
		"dashboard.grafana.app/folders:update",
	}

	group := "dashboard.grafana.app"
	resource := "dashboards"
	verb := "get"
	name := "dashUID"

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if !hasPermissionInToken(tokenPermissions, group, resource, verb, name) {
			b.Fail()
		}
	}
}

func BenchmarkHasPermissionInToken_Prefixs(b *testing.B) {
	tokenPermissions := []string{
		"dashboard.grafana.app:get",
	}

	group := "dashboard.grafana.app"
	resource := "dashboards"
	verb := "get"
	name := "dashUID"

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if !hasPermissionInToken(tokenPermissions, group, resource, verb, name) {
			b.Fail()
		}
	}
}
```

</details>

### DIFF
--- a/authz/client.go
+++ b/authz/client.go
@@ -6,6 +6,7 @@ import (
 	"encoding/gob"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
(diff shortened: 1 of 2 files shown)


IevaVasiljeva (approved):
🚢

authz/client.go
```diff
@@ -292,6 +293,37 @@ func (c *ClientImpl) check(ctx context.Context, id claims.AuthInfo, req *CheckRe
 	return resp.Allowed, err
 }
 
+func hasPermissionInToken(tokenPermissions []string, group, resource, verb, name string) bool {
```
eleijonmarck, line 296: nice work on optimizations!

is there a way to describe the algorithm, and what makes it so fast?

// check second group part first
// group cannot be verb
// check first group part second
// requires a grouping
// grouping has prefix

eleijonmarck (approved):
nit: on comment on description on why the algorithm is fast

prefix check, smells alot like a trie

authz/client_test.go
```diff
@@ -14,6 +14,118 @@ import (
...
+			group:            "dashboard.grafana.app",
+			resource:         "dashboards",
+			verb:             "list",
+			want:             true,
+		},
+		{
+			name:             "Wildcard action works",
+			tokenPermissions: []string{"dashboard.grafana.app/dashboard:*"},
+			group:            "dashboard.grafana.app",
+			resource:         "dashboards",
```
cinaglia, line 114: I'm lacking quite a bit of context here, but is this the intended behavior? IMO it seems dangerous to implement something like this. For instance, if we were to ever introduce another resource called `dashboard-annotations` (as a crude example), the token would automatically have access over it.

```suggestion
			tokenPermissions: []string{"dashboard.grafana.app/dashboards:*"},
			group:            "dashboard.grafana.app",
			resource:         "dashboards",
```

gamab: Excellent point thanks! I'll fix that tomorrow

chaudyg (approved):

authz/client.go
gamab: Is the new impl simpler to read?

PR #125, opened 2024-12-19
Authn: Fix validation check for id tokens

kalleep:
Fix issue when a request contains a id token for anything other than user or service account


gamab (approved):

gamab:
As discussed, we'll need to make sure the AuthZ Client and Server do an identity check

PR #127, opened 2024-12-20
AuthZ: Implement Compile

gamab:
Implement compile.

- [x] Review my code
- [x] Add more tests
- [ ] I don't really understand why we take the namespace given it's provided when you call compile.
- [x] I wonder if we should account for the granular permission in token case as it makes it harder to build the ItemChecker.
- [x] We should cache the list response.


IevaVasiljeva (approved):
Sweet, good to see `List` being used so soon.

authz/client.go
```diff
@@ -389,6 +370,119 @@ func (c *ClientImpl) Check(ctx context.Context, id claims.AuthInfo, req CheckReq
...
+	span.SetAttributes(attribute.Bool("with_user", !isService))
+
+	// No user => check on the service permissions
+	if isService {
+		// access token check is disabled => we can skip the authz service
+		if !c.authCfg.accessTokenAuthEnabled {
+			return allowAllChecker(list.Namespace), nil
+		}
+
+		// ToDo - should we handle the granular permission case?
```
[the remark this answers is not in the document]
gamab: As discussed during the daily, we won't support granular permissions anymore until someone explicitly requests it/

authz/client.go
```diff
@@ -389,6 +370,117 @@ func (c *ClientImpl) Check(ctx context.Context, id claims.AuthInfo, req CheckReq
...
+			return denyAllChecker, nil
+		}
+	}
+
+	// Instantiate a new context for the request
+	outCtx := newOutgoingContext(ctx)
+
+	// Query the authz service
+	checkReq := &authzv1.ListRequest{
+		Subject:   id.GetSubject(),
```
gamab, line 423: Missing Verb

authz/client.go
```diff
@@ -385,6 +366,137 @@ func (c *ClientImpl) Check(ctx context.Context, id claims.AuthInfo, req CheckReq
...
+	outCtx := newOutgoingContext(ctx)
+
+	// Query the authz service
+	listReq := &authzv1.ListRequest{
+		Subject:     id.GetSubject(),
+		Group:       list.Group,
+		Resource:    list.Resource,
+		Verb:        list.Verb,
+		Namespace:   list.Namespace,
+		Subresource: list.Subresource,
```
IevaVasiljeva, line 386: What is the subresource?

authz/client.go
```diff
@@ -385,6 +366,137 @@ func (c *ClientImpl) Check(ctx context.Context, id claims.AuthInfo, req CheckReq
...
+		return nil, err
+	}
+
+	if !c.validateCallerNamespace(id, list.Namespace) {
+		return denyAllChecker, nil
+	}
+
+	span.SetAttributes(attribute.String("namespace", list.Namespace))
+	span.SetAttributes(attribute.String("group", list.Group))
+	span.SetAttributes(attribute.String("resource", list.Resource))
```
IevaVasiljeva, line 420: nit: could also add attribute for the verb.

IevaVasiljeva (approved):
Great work! LGTM with two minor questions/comments.

authz/client.go
gamab: Should :D

authz/client.go
gamab: We don't have any yet. Just being future proof there. 
I think it could be something like teams/<uid>/members/

PR #130, opened 2025-01-14
Chore: Add json tags to Claims

mgyongyosi:


kalleep (commented):
Would it now work to have inline tags for `jwt.Claims` and `Rest`?

mgyongyosi:
> Would it now work to have inline tags for `jwt.Claims` and `Rest`?

I tried, but `inline` does not supported by the `encoding/json` package. :/ https://github.com/golang/go/issues/6213

kalleep (approved):

gamab (approved):

PR #131, opened 2025-01-17
AccessClient: Move the interface to a new types package

ryantxu:
Can we move the access client back to a simple interface package with limited dependencies?

Having it in the authz package requires consumers to depend on the full dependency tree.  eg:
https://github.com/grafana/grafana/blob/main/pkg/apimachinery/go.mod#L14 😢 

I also think it will be more clear if we can keep the interface separate and simple


gamab (approved):

gamab:
Conceptually, I liked having the `AccessClient` in the `authz` package. Having it in the claims package conveys the message that the `AccessClient` interface works with claims.
That being said I understand your point of having a lighter dependency tree.

kalleep (approved):
Discussed in slack, a more generic name is better for this package than claims. I am ok with `typed`. Any opinions on this @gamab ?

ryantxu:
See: https://github.com/grafana/grafana/pull/99243

PR #134, opened 2025-01-23
AuthZ: Quick fix for the service on behalf of a service flow

gamab:
When we use a cap token to connect to a grafana instance, an ID token is generated.

When the grafana instance makes a request to the resource store, this ID token is shipped along with Grafana's access token.

Upon checking permissions in the AuthZ client, the call is correctly flagged as "service call" but given the id token presence, the permissions associated is empty.

This PR intends to be a quick fix for this until we have the first service's permissions in the ID token.


authn/auth_info.go
```diff
@@ -93,7 +93,11 @@ func (a *AuthInfo) GetAuthenticatedBy() string {
 }
 
 func (a *AuthInfo) GetTokenPermissions() []string {
-	if a.id != nil {
+	// If it's a service acting on behalf of a user
+	// we should not check token permission but delegated permissions instead
+	// If it's a service acting on behalf of a second service
+	// we currently just check the first service permissions
+	if a.id != nil && a.id.Rest.Type != types.TypeAccessPolicy {
 		return []string{}
```
gamab, line 101: @cinaglia pointed out to me that this should probably be an error case.
If we are checking tokenPermissions when it's a user, we are doing something wrong.

kalleep (approved):

authn/auth_info.go
cinaglia: Ideally, the interface for ID tokens wouldn't even have functions it doesn't support (i.e. `GetTokenPermissions`), but I understand that would over complicate things. In the absence of that, I think returning a descriptive error for the invalid scenario would be very beneficial.

PR #138, opened 2025-01-30
feat: add org role claim to id token

charandas:
We have had a need come up in addressing a need in the new Direct API Server access project in app platform, which requires being able to determine user's access in cloud auth webhook with a single Authorization header.

The issue is that the authz-service currently is not capable for helping cloud webhook determine what the user's org role is. Until that is the case, we would like to package that within the id token and use our own proprietary format to string together the tokens in the Authorization header.

### DIFF
--- a/authn/verifier_id_token.go
+++ b/authn/verifier_id_token.go
@@ -23,8 +23,8 @@ type IDTokenClaims struct {
 	Username string `json:"username,omitempty"`
 	// Display name of the user (name attribute if it is set, otherwise the login or email)
 	DisplayName string `json:"name,omitempty"`
-	// User's role in Grafana Org
-	OrgRole string `json:"orgRole,omitempty"`
+	// Basic role of entity (Viewer, Editor, Admin)
+	Role string `json:"role,omitempty"`
 }
 
 // Helper for the id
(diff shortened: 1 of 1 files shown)


kalleep:
I re-opened https://github.com/grafana/identity-access-team/issues/692

I can make the pr in grafana to start signing tokens with the role so I can make sure we invalidate the id token cache for entities whenever the role changes.

authn/verifier_id_token.go
```diff
@@ -23,6 +23,8 @@ type IDTokenClaims struct {
 	Username string `json:"username,omitempty"`
 	// Display name of the user (name attribute if it is set, otherwise the login or email)
 	DisplayName string `json:"name,omitempty"`
+	// User's role in Grafana Org
+	OrgRole string `json:"orgRole,omitempty"`
```
kalleep, line 27:
```suggestion
	// Basic role of entity (Viewer, Editor, Admin)
	Role string `json:"role,omitempty"`
```

cinaglia:
@kalleep do we need to do anything in the Auth API for this? It seems like we might but I can't be certain.

kalleep:
@cinaglia no we don't, we just need to populate the claim in grafana and it will be sent in "extra". Auth api will start adding that claim to new tokens

PR #143, opened 2025-02-14
Folders: Hide k6 folders for users and show them for SAs

IevaVasiljeva:
**What**

Don't allow users to see or list k6 folders, but allow service accounts to access them.

This mirrors the work that we did previously to prevent users from accidentally deleting k6 folders, which are used to scope k6 projects and their permissions.

Previous PRs on the legacy code paths: https://github.com/grafana/grafana/pull/88884 and https://github.com/grafana/grafana/pull/90008

Note: I did initially want to do this in the code that lists folders, but it required too many changes to the search code, so this is simpler and probably also safer.

### DIFF
--- a/authz/client.go
+++ b/authz/client.go
@@ -165,6 +165,11 @@ func (c *ClientImpl) check(ctx context.Context, id types.AuthInfo, req *types.Ch
 	ctx, span := c.tracer.Start(ctx, "ClientImpl.hasAccess")
 	defer span.End()
 
+	idIsServiceAccount := types.IsIdentityType(id.GetIdentityType(), types.TypeServiceAccount)
+	if !idIsServiceAccount && (req.Name == "k6-app" || req.Folder == "k6-app") {
+		return false, nil
+	}
+
 	key := checkCacheKey(id.GetSubject(), req)
 	res, err := c.getCachedCheck(ctx, key)
 	if err == nil {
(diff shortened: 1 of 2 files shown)


authz/client.go
```diff
@@ -165,6 +165,11 @@ func (c *ClientImpl) check(ctx context.Context, id types.AuthInfo, req *types.Ch
 	ctx, span := c.tracer.Start(ctx, "ClientImpl.hasAccess")
 	defer span.End()
 
+	idIsServiceAccount := types.IsIdentityType(id.GetIdentityType(), types.TypeServiceAccount)
+	if !idIsServiceAccount && (req.Name == "k6-app" || req.Folder == "k6-app") {
+		return false, nil
+	}
```
kalleep, line 171: I know we have this hack in the old code to solve something, but this is not a great solution.. 

Is there any plan to get rid of this hack? I would be more comfortable to add the hack if we have a way to get rid of it.

IevaVasiljeva: As we start storing more and more things in folders, we will need to redesign the UI to account for it. The main issue is that currently we only show dashboards stored in folders under the folder tree, so users don't know what else a folder is used for. We discussed this with the app platform team, and everyone agrees that we'll need to expand the folder view. No concrete plans yet though.

eleijonmarck (approved):
tested locally and works perfectly ⚡

PR #150, opened 2025-02-19
validate Check/Compile requests

chaudyg:
Ensure we validate the request consistently across implementations of Check/Compile.

Context [thread](https://raintank-corp.slack.com/archives/C06JFJ7H3GC/p1739975483455959)

### DIFF
--- a/authz/client.go
+++ b/authz/client.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/gob"
-	"errors"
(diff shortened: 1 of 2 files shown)


eleijonmarck (approved):
Approving, but not sure why this makes it consistent. Is it consistent across check and list we want?

Confused perhaps on the discussion of https vs grpc, which is not related to this

gamab (approved):

Issue #152, opened 2025-02-21
Atuz: Compile / list should not require that namespace is set if caller has namespace '*'


kalleep:
Right now we validate in client code that a `ListRequest` has a namespace set. 

This is not entirely correct because it makes it impossible to watch resources across namespaces, something e.g. an operator can do.

This is a bug that was detected in https://github.com/grafana/grafana/pull/101124.

So we need to fix that check in the case the caller has access to namespace `*` and that can only be true for * `types.TypeAccessPolicy`.

We should also remove namespace from [ItemChecker](https://github.com/grafana/authlib/blob/main/types/access.go#L79)

PR #153, opened 2025-02-21
Authz: namespace validation

kalleep:
In https://github.com/grafana/grafana/pull/101124 we discovered that it is impossible to use watch across namespaces and this is considered a bug.

Watching for a resource type across namespaces will happen whenever we deploy multi-tenant operators.

I changed the validation checks to never check for empty namespaces and moved the namespace matches check into validation instead so its done both for `FixedAccessClient` and authz implementation. This is safe because only `TypeAccessPolicy` will ever be able to perform this kind of watch. Any other type is forwarded to authz service where an empty namespace woul always be considered errors. Also there _should_ never be a case where we get an identity without the namespace set for it so it _should_  never match an empty namespace. It is also only possible for `TypeAccessPolicy` to have wildcard namespace and that is the only way to match it.

Fixes https://github.com/grafana/authlib/issues/152


types/namespace_test.go
```diff
@@ -1,130 +1,182 @@
...
+package types
 
 import (
 	"testing"
 
-	"github.com/grafana/authlib/types"
+	"github.com/stretchr/testify/assert"
 )
 
+func TestNamespaceMatches(t *testing.T) {
```
kalleep, line 9: Added missing tests

authz/client.go
```diff
@@ -252,18 +251,14 @@ func (c *ClientImpl) Compile(ctx context.Context, id types.AuthInfo, list types.
...
 	if id.GetSubject() == "" {
-		span.RecordError(types.ErrMissingCaller)
-		return nil, types.ErrMissingCaller
-	}
-
-	if !types.NamespaceMatches(id.GetNamespace(), list.Namespace) {
-		return denyAllChecker, nil
+		span.RecordError(ErrMissingAuthInfo)
+		return nil, ErrMissingAuthInfo
 	}
```
kalleep, line 262: Before we returned a `denyAllChecker` if the namespaces did not match, this seems really wrong because the request would not be short circuited.

The implication of returning a `denyAllChecker` is:
1. For list we would paginate through all resources and always end up with an empty result
2. Watch would start but we would always filter out every event

So the correct thing must be to return an error here.

gamab: Really good point!

types/namespace_test.go
```diff
@@ -1,130 +1,182 @@
...
+			namespace:         "*",
+			expectedNamespace: "stack-1",
+			expected:          true,
+		},
+		{
+			desc:              "wildcard namespace and empty expected namespace",
+			namespace:         "*",
+			expectedNamespace: "",
+			expected:          true,
+		},
```
gamab, line 45: Self note: To look into

authz/client.go
```diff
@@ -154,18 +160,18 @@ func (c *ClientImpl) Check(ctx context.Context, id types.AuthInfo, req types.Che
...
 	if id.GetSubject() == "" {
-		span.RecordError(types.ErrMissingCaller)
-		return checkResponseDenied, types.ErrMissingCaller
+		span.RecordError(ErrMissingAuthInfo)
+		return checkResponseDenied, ErrMissingAuthInfo
 	}
 
 	if !types.NamespaceMatches(id.GetNamespace(), req.Namespace) {
-		return checkResponseDenied, nil
+		return checkResponseDenied, namespaceMissmatchError(id.GetNamespace(), req.Namespace)
```
kalleep, line 174: I moved this check back to authz client and no longer perform it in fixed client.

A fixed client should always give whatever result is configured and with this check there it would not be true, breaking what is expected for that client implementation. cc @gamab @chaudyg

gamab (commented):
This *should* not happen, but we should still add a check that the caller namespace is not empty otherwise an access policy without a namespace would have access to all namespaces.

types/namespace.go
```diff
@@ -28,14 +28,13 @@ func disambiguateNamespace(namespace string) string {
...
-	// actual should never be a "*" where ID token claims are concerned
-	if actual == "*" {
+// NamespaceMatches check if provided namespace matches the expected one.
+// This function always cosider the namespace to match if namespace is `*`.
+func NamespaceMatches(namespace, expected string) bool {
+	if namespace == "*" {
 		return true
 	}
-	return actual == expected
+	return disambiguateNamespace(namespace) == disambiguateNamespace(expected)
```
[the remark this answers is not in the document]
kalleep: Yes I agree, you made a comment before that was deleted. I wrote something about that.

I think the validation should be that the only thing that can match empty namespace is the wildcard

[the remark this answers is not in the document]
kalleep: [4595428](https://github.com/grafana/authlib/pull/153/commits/4595428b57760cbee123831de7e48fa0b3399de6)

PR #155, opened 2025-02-26
Authz: We don't use "list" for authz so we should not require it for access tokens

kalleep:
We always map verb `list` to `get` for authorization. But here we do required it for access tokens. So if you want to list using access token you also need `group/resource:list`. We should re-map it for them too

### DIFF
--- a/authz/client.go
+++ b/authz/client.go
@@ -131,6 +131,11 @@ func (c *ClientImpl) check(ctx context.Context, id types.AuthInfo, req *types.Ch
 }
 
 func hasPermissionInToken(tokenPermissions []string, group, resource, verb string) bool {
+	// we always map list to get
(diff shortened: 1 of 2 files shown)


authz/client.go
```diff
@@ -131,6 +131,11 @@ func (c *ClientImpl) check(ctx context.Context, id types.AuthInfo, req *types.Ch
 }
 
 func hasPermissionInToken(tokenPermissions []string, group, resource, verb string) bool {
+	// we always map list to get
+	if verb == "list" {
+		verb = "get"
+	}
```
kalleep, line 137: I wonder if we should accept both to be backwards compatible, wdyt @gamab

IevaVasiljeva (approved):

PR #156, opened 2025-03-04
Authz: Add error helpers

gamab:
This pull request introduces new set of helper functions to handle error checking in the `authz/client.go` file. These changes improve error handling by providing specific functions to identify unauthenticated and unauthorized errors.

### DIFF
--- a/authz/client.go
+++ b/authz/client.go
@@ -31,6 +31,14 @@ var (
 	ErrNamespaceMissmatch = errors.New("namespace missmatch")
 )
 
+func IsUnauthenticatedErr(err error) bool {
+	return errors.Is(err, ErrMissingAuthInfo)
+}
+
+func IsUnauthorizedErr(err error) bool {
+	return errors.Is(err, ErrNamespaceMissmatch)
+}
+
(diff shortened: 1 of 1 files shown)


authz/client.go
```diff
@@ -31,6 +31,14 @@ var (
 	ErrNamespaceMissmatch = errors.New("namespace missmatch")
 )
 
+func IsUnauthenticatedErr(err error) bool {
+	return errors.Is(err, ErrMissingAuthInfo)
+}
```
kalleep, line 36: This is a bit unfortunate https://github.com/grafana/authlib/blob/main/authn/errors.go#L26. 

I think we can skip `IsUnauthenticatedErr`. I think if we get here without any auth info something is really wrong.

gamab: Why unfortunate when it's consistent?

> I think we can skip IsUnauthenticatedErr. I think if we get here without any auth info something is really wrong.

👌

kalleep (approved):

IevaVasiljeva (approved):

PR #157, opened 2025-03-04
Docs: Update documentation

kalleep:
Instead of having docs in REAME I create a docs folder. In this folder I added `authorization` and `authentication` where I put what we currently have. We should extend the docs with examples for other scenarios but I will leave that for another pr.

I remove the example using `inproc` because at least right now that is highly specific to grafana.


docs/authentication/grpc.md
```diff
@@ -0,0 +1,132 @@
+# Authentication using GRPC
+
+To perform authentication between two services using access and id tokens we need to setup a grpc compatible authenticator on for the server and have a way for the client to generate the access token.
+
+### Example server side
+
+For a grpc server we can leverage our generic [`Authenticator`](../authn/authenticator.go). To set it up we need to configure a [`KeyRetriever`](../authn/jwks.go) and setup [`AccessTokenVerifier`](../authn/verifier_access_token.go) and
```
IevaVasiljeva, line 7:
```suggestion
For a grpc server we can leverage our generic [`Authenticator`](../authn/authenticator.go). To set it up we need to configure a [`KeyRetriever`](../authn/jwks.go) and set up [`AccessTokenVerifier`](../authn/verifier_access_token.go) and  
```

docs/authentication/grpc.md
```diff
@@ -0,0 +1,132 @@
+# Authentication using GRPC
+
+To perform authentication between two services using access and id tokens we need to setup a grpc compatible authenticator on for the server and have a way for the client to generate the access token.
```
IevaVasiljeva, line 3:
```suggestion
To perform authentication between two services using access and id tokens we need to setup a grpc compatible authenticator for the server and have a way for the client to generate the access token.
```

IevaVasiljeva (commented):
Would it make sense to also move `authz/README.md` content to `docs/authorization`?

kalleep:
> Would it make sense to also move `authz/README.md` content to `docs/authorization`?

Yes it does. I moved it into `single-tenant` and remove what we don't need to have there

eleijonmarck (approved):
Nice cleaning. Nit on removing the inproc, we should add that doc to the grafana document then

PR #160, opened 2025-03-11
AuthZ: Fix split count in `hasPermissionInToken`

gamab:
This PR fixes the split count in `hastPermissionInToken` since we expect at max 2 parts.
This currently has no incidence given the explicit switch case statement was already filtering out `>3` parts results.


kalleep (approved):
Oh yeah because previously we supported names right

PR #163, opened 2025-03-14
Add grpc pkg to authlib

leonorfmartins:
Context: https://github.com/grafana/search-and-storage-team/issues/220

Testing this with a local replace of this module in the go.mod file like this:
```
replace github.com/grafana/authlib => <local repo path>/authlib
```


kalleep:
We just updated a bunch of stuff in go.* maybe just checkout what is on main. This pr don't have to update anything in those files

grpcutils/config.go
```diff
@@ -0,0 +1,8 @@
+package grpcutils
+
+type GrpcAuthenticatorConfig struct {
```
kalleep, line 3:
```suggestion
type AuthenticatorConfig struct {
```
I think the grpc part can be inferred from that package name

```
grpcutils.GrpcAuthenticatorConfig{}
```
vs 
```
grpcutils.AuthenticatorConfig{}
```

grpcutils/config.go
```diff
@@ -0,0 +1,8 @@
...
+	SigningKeysURL   string
+	AllowedAudiences []string
+	LegacyFallback   bool
```
kalleep, line 6:
```suggestion

```
Not needed here

grpcutils/grpc_authenticator.go
```diff
@@ -0,0 +1,82 @@
...
+	Authenticate(ctx context.Context) (context.Context, error)
+}
+
+type GAuthenticatorFunc func(context.Context) (context.Context, error)
+
+func (fn GAuthenticatorFunc) Authenticate(ctx context.Context) (context.Context, error) {
+	return fn(ctx)
+}
+
+func NewUnsafeAuthenticator() grpcAuthenticator {
```
kalleep, line 29:
```suggestion
func NewUnsafeAuthenticator(tracer trace.Tracer) grpcAuthenticator {
```
Caller can choose to pas a noop tracer or a real one I guess

grpcutils/grpc_authenticator.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/grafana/authlib/types"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+type AuthenticatorConfig interface {
```
gamab, line 19: Nit.
```suggestion
type Authenticator interface {
```

grpcutils/grpc_authenticator.go
```diff
@@ -0,0 +1,82 @@
...
+	Authenticate(ctx context.Context) (context.Context, error)
+}
+
+type GAuthenticatorFunc func(context.Context) (context.Context, error)
```
gamab, line 23: Do you define it not to import?
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/auth/auth.go#L24

grpcutils/grpc_authenticator.go
```diff
@@ -0,0 +1,82 @@
...
+
+type grpcAuthenticator interface {
+	Authenticate(ctx context.Context) (context.Context, error)
+}
+
+type GAuthenticatorFunc func(context.Context) (context.Context, error)
+
+func (fn GAuthenticatorFunc) Authenticate(ctx context.Context) (context.Context, error) {
+	return fn(ctx)
+}
```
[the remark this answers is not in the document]
leonorfmartins: you're right they're equivalent but I'd rather keep the type for 2 reasons:
- we still use it in Grafana in different places and I'd rather not change too many things at once
- I find it more readable as in it tells us better what it does (I understand how this is more of a prefence)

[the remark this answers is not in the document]
kalleep: Sure I get that but we don't really have a need for this interface. I guess we reused it to be able to return different types here https://github.com/grafana/grafana/blob/main/pkg/services/authn/grpcutils/grpc_authenticator.go#L53 but that is not really needed.

So lets not expose an interface that we don't use and just make every `New*` return func(context.Context) (context.Context, error)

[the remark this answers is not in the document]
leonorfmartins: got it :) if this is the only use case, I can convert it.

go.work
```diff
@@ -1,3 +1,4 @@
+
```
kalleep, line 1:
```suggestion
```

grpcutils/grpc_authenticator.go
```diff
@@ -0,0 +1,78 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+type AuthenticatorFunc func(context.Context) (context.Context, error)
+
+func (fn AuthenticatorFunc) Authenticate(ctx context.Context) (context.Context, error) {
+	return fn(ctx)
+}
```
kalleep, line 23:
```suggestion
```

kalleep: We don't need this either :)

grpcutils/config.go
kalleep: We should still rename this one

kalleep (commented):
Just some more things then we are good to go :+1:

grpcutils/grpc_authenticator.go
leonorfmartins: ah true! I removed `AuthenticatorFunc ` from the `New*` return then :)

grpcutils/grpc_authenticator.go
```diff
@@ -0,0 +1,72 @@
...
+	"google.golang.org/grpc/status"
+)
+
+func NewUnsafeAuthenticator(tracer trace.Tracer) func(ctx context.Context) (context.Context, error) {
+	return NewAuthenticatorInterceptor(
+		authn.NewDefaultAuthenticator(
+			authn.NewUnsafeAccessTokenVerifier(authn.VerifierConfig{}),
+			authn.NewUnsafeIDTokenVerifier(authn.VerifierConfig{}),
+		),
+		noop.NewTracerProvider().Tracer(""),
```
gamab, line 25:
```suggestion
		tracer,
```

gamab (approved):
LGTM :)

kalleep (approved):
LGTM, with @gamab suggestion

PR #164, opened 2025-03-21
[WIP] validation of interface

eleijonmarck:


eleijonmarck:
not needed, need to be checked in the repo where we get the authinfo

PR #165, opened 2025-03-24
Add Actor to AccessTokenClaims

mgyongyosi:
Add Actor ("act") claim to AccessTokenClaims.

### DIFF
--- a/authn/verifier_access_token.go
+++ b/authn/verifier_access_token.go
@@ -5,10 +5,8 @@ import (
 )
 
 type ActorClaims struct {
-	Subject              string       `json:"sub"`
-	Permissions          []string     `json:"permissions,omitempty"`
-	DelegatedPermissions []string     `json:"delegated_permissions,omitempty"`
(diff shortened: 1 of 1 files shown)


authn/verifier_access_token.go
```diff
@@ -4,6 +4,13 @@ import (
 	"context"
 )
 
+type ActorClaims struct {
+	Subject              string       `json:"sub"`
+	Permissions          []string     `json:"permissions,omitempty"`
+	DelegatedPermissions []string     `json:"delegated_permissions,omitempty"`
```
gamab, line 10: I had understood we'd be overriding the permissions and delegated permissions of the Access Token with the intersect (to keep the AuthZ client unchanged)
Did I misunderstand?

mgyongyosi: You did not misunderstand, the intersection is in the access token! So, you mean that the "original" caller's permission won't be necessary to be sent in the "act" claim, right?

gamab (approved):

PR #166, opened 2025-04-02
Add WithCache option to for NewKeyRetriever

mem:
It might be worth offering the option to set a custom cache, possibly with different policies or with a different caching store.

This is almost supported, as cache.Cache is an interface, so it's possible to reimplement it. This is done using a
DefaultKeyRetrieverOption function to pass to NewKeyRetriever that can replace the cache field `c`. The problem is that the field is not exported, so it's not possible to implement the option outside of the package.

Add that option here. It needs to take into consideration the fact that cache.NewLocalCache ends up starting a goroutine to do clean up tasks. While the goroutine should stop because the cache has been replaced, it's better to not start it in the first place.

### DIFF
--- a/authn/jwks_test.go
+++ b/authn/jwks_test.go
@@ -3,12 +3,16 @@ package authn
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/go-jose/go-jose/v3"
+	"github.com/grafana/authlib/cache"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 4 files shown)


authn/jwks_test.go
```diff
@@ -66,3 +72,91 @@ func TestDefaultKeyRetriever_Get(t *testing.T) {
...
+	// The cache is empty, so the implementation should fetch the key.
+	key, err := service.Get(context.Background(), firstKeyID)
+	require.NoError(t, err)
+	require.NotNil(t, key)
+	assert.Equal(t, firstKeyID, key.KeyID)
+
+	// If the implementation called the cache, the data should be there.
+	data, err := tc.Get(context.Background(), firstKeyID)
+	require.NoError(t, err)
+	require.NotNil(t, data)
```
mem, line 106: I couldn't figure out a way to validate that this is the right data.

cinaglia (approved):
LGTM

colin-stuart (approved):

authn/jwks_test.go
mem: OK, thanks for taking a look and thank you for the suggestion!

mem: I did apply that.

PR #168, opened 2025-04-09
AuthZ: Switch to uids

gamab:
Zanzana is using user UIDs to perform access control checks.
The AuthZ service is able to work with both IDs and UIDs.
This PR proposes a switch to UIDs.

We have to check for other identity kinds that everything remains operational. (Anonymous, Rendering)


IevaVasiljeva (approved):
I tested both the renderer and anonymous flows, and they both seem to work fine. We only care about the correct identity type being set for these flows ([code](https://github.com/grafana/grafana/blob/v11.6.0/pkg/services/authz/rbac/service.go#L281-L284)), so it should work fine with both ID (0) or UID (an empty string).

PR #173, opened 2025-05-13
Allow user as actor in access token

Rydez:
Part of https://docs.google.com/document/d/1jPsKAsobIXPnVZgllNZGfIMqFA8kwVGfq92H6pmrDoY/edit?tab=t.0.

@mgyongyosi I'm definitely a bit naive on some of this stuff, but I figured I'd get a PR up so we could work async a bit more easily. Wdyt?

I've started the Auth API work as well, but since it depends on this I figured I'd start here.

### DIFF
--- a/authn/auth_info.go
+++ b/authn/auth_info.go
@@ -14,12 +14,32 @@ type AuthInfo struct {
 }
 
 func NewAccessTokenAuthInfo(at Claims[AccessTokenClaims]) *AuthInfo {
+	var id *Claims[IDTokenClaims]
+	actorIdentity := at.Rest.getIdentity()
+	if actorIdentity != nil {
+		id = &Claims[IDTokenClaims]{
+			Rest:  *actorIdentity,
+			token: at.token,
+		}
+	}
+
 	return &AuthInfo{
 		at: at,
+		id: id,
 	}
 }
 
 func NewIDTokenAuthInfo(at Claims[AccessTokenClaims], id *Claims[IDTokenClaims]) *AuthInfo {
+	if id == nil {
+		actorIdentity := at.Rest.getIdentity()
+		if actorIdentity != nil {
+			id = &Claims[IDTokenClaims]{
+				Rest:  *actorIdentity,
+				token: at.token,
+			}
+		}
+	}
+
 	return &AuthInfo{
 		at: at,
 		id: id,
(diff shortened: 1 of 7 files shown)


authz/client.go
```diff
@@ -93,23 +93,23 @@ func NewClient(cc grpc.ClientConnInterface, opts ...AuthzClientOption) *ClientIm
 // Implementation
 // -----
 
-func (c *ClientImpl) check(ctx context.Context, id types.AuthInfo, req *types.CheckRequest) (bool, error) {
```
Rydez, line 96: Sorry for the noise just to rename a variable, but this was confusing me at first.

authn/verifier_access_token.go
```diff
@@ -25,6 +29,34 @@ type AccessTokenClaims struct {
...
+	currentActor := c.Actor
+	if currentActor != nil {
+		for currentActor.Actor != nil {
+			currentActor = currentActor.Actor
+		}
+	}
+
+	return currentActor
+}
+
```
cinaglia, line 59: I think we can simplify things a bit if we introduced a function to get the identity actor that builds on `getInnermostActor`. It can then check the identity and return nil if it it's not of type `User`.

```suggestion

func (c AccessTokenClaims) getIdentity() *IDTokenClaims {
	actor := c.getInnermostActor()
	if actor != nil && types.IsIdentityType(actor.IDTokenClaims.Type, types.TypeUser) {
		return &actor.IDTokenClaims
	}

	return nil
}
```

authn/auth_info.go
cinaglia, line 1: We might be able to simplify this quite a bit by updating `NewAccessTokenAuthInfo` and `NewIDTokenAuthInfo` to set `id` accordingly based on whether `at.Rest.getIdentity()` returns a non-nil object. Something along the lines of the following:

```go
func NewIDTokenAuthInfo(at Claims[AccessTokenClaims], id *Claims[IDTokenClaims]) *AuthInfo {
	if id == nil {
		actorIdentity := at.Rest.getIdentity()
		if actorIdentity != nil {
			id = &Claims[IDTokenClaims]{
				Rest:  *actorIdentity,
				token: at.token,
			}
		}

	}

	return &AuthInfo{
		at: at,
		id: id,
	}
}
```

That would make it so we can re-use the `IDTokenClaims.getK8sName()` and `IDTokenClaims.getTypedUID()` implementations, and ultimately remove all the extra handling in these getters (with the exception of `AuthInfo.GetSubject()` which should still be built on top of `getInnermostActor()`.

authn/verifier_access_token.go
Rydez: Oh nice. Are you suggesting to use `getIdentity` instead of `getInnermostActor` in the `auth_info.go` functions?

authn/auth_info.go
Rydez: Ah, yeah, that makes sense. Good call. I'll put together a commit.

Rydez: Handled in https://github.com/grafana/authlib/pull/173/commits/5db6cb3dc919f0508cd9b9909b2bdb68251eaeba.

IevaVasiljeva (commented):
Looks good from what I can tell, just left two questions/comments. Also curious about @gamab's thoughts on this.

authn/auth_info.go
```diff
@@ -148,5 +168,5 @@ func (a *AuthInfo) GetIDToken() string {
 	if a.id != nil {
 		return a.id.token
 	}
-	return ""
+	return a.at.token
```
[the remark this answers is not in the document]
Rydez: Ah, yeah, I'll remove this change. `at.token` is set to the `id.token` in `getIdInfo()`.

gamab (approved):

gamab:
Great job 💯

authn/auth_info.go
```diff
@@ -14,37 +15,60 @@ type AuthInfo struct {
...
+// This can be the case if an ID token is sent in the request to sign an access token.
+func getIdInfo(at Claims[AccessTokenClaims]) *Claims[IDTokenClaims] {
+	identityActor := at.Rest.getIdentityActor()
+	if identityActor == nil {
+		return nil
+	}
+
+	return &Claims[IDTokenClaims]{
+		Rest:  identityActor.IDTokenClaims,
+		token: at.token,
```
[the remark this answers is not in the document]
Rydez: I also see that `GetIDToken()` uses this.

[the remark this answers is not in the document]
gamab: Ah 🤔 
Let's maybe not set it then as this is technically an access-token (an ID token does not grant any access)

PR #175, opened 2025-05-14
Return serviceIdentity in the AuthInfo.GetExtra if it's not empty

mgyongyosi:
Return the `AccessTokenClaims.Rest.ServiceIdentity` in `AuthInfo.GetExtra()` if the service identity is specified/not empty.

### DIFF
--- a/authn/const.go
+++ b/authn/const.go
@@ -6,4 +6,6 @@ const (
 
 	httpHeaderAccessToken = "X-Access-Token"
 	httpHeaderIDToken     = "X-Grafana-Id"
+
+	ServiceIdentityKey = "serviceIdentity"
 )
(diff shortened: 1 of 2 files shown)


macabu (approved):
LGTM, thank you!

authn/const.go
```diff
@@ -6,4 +6,6 @@ const (
 
 	httpHeaderAccessToken = "X-Access-Token"
 	httpHeaderIDToken     = "X-Grafana-Id"
+
+	ServiceIdentityKey = "serviceIdentity"
```
gamab, line 10: Nit. Should we be a bit more precise?
```suggestion
	ServiceNameExtraKey = "serviceIdentity"
```

gamab, line 10: Or
```suggestion
	ExtraKeyServiceIdentity = "serviceIdentity"
```

PR #176, opened 2025-06-03
feat: Add WithMinimumCacheTTL option

mem:
The WithMinimumCacheTTL option is used to control when cache entries are removed from the cache. The existing code sets this value to 15 seconds, and if this option is not passed, that is the fallback value.

### DIFF
--- a/authn/token_exchange.go
+++ b/authn/token_exchange.go
@@ -22,6 +22,8 @@ type TokenExchanger interface {
 	Exchange(ctx context.Context, r TokenExchangeRequest) (*TokenExchangeResponse, error)
 }
 
+const defaultCacheTTL = 15 * time.Second
+
 var _ TokenExchanger = &TokenExchangeClient{}
 
 // ExchangeClientOpts allows setting custom parameters during construction.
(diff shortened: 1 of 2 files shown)


authn/token_exchange.go
```diff
@@ -40,6 +42,14 @@ func WithTokenExchangeClientCache(cache cache.Cache) ExchangeClientOpts {
 	}
 }
 
+// WithMinimumCacheTTL allows setting the minimum amount of time that a cache
+// entry must be valid for in order for it to be reused.
+func WithMinimumCacheTTL(ttl time.Duration) ExchangeClientOpts {
```
cinaglia, line 47: I'm curious to hear your use case for making this configurable. The 15-second leeway acts as way to enforce that a token is purged from the cache a few seconds before its actual expiration. Would you actually like it to be purged sooner/later?

cinaglia: I spoke to @mem offline and got some context around this. I understand the reasoning behind it and think it's a reasonable change.

I did share some minor concern about the naming here as `WithMinimumCacheTTL` doesn't quite convey what this is meant to achieve. That said, I'm not sure I can provide a substantially better alternative. Some ideas:

- WithExpirationBuffer
- WithEvictionThreshold

cinaglia (approved):
Functionality-wise this looks good to me. I understand the need for it.

My only suggestions here would be to rename the option to something hopefully more self-explanatory. I'd also suggest providing more context around why this is needed, perhaps by providing a real-life example.

cinaglia:
@mem is this still something you'd like to wrap up?

PR #177, opened 2025-06-10
AuthN: Maintain service identity during obo requests

gamab:
This pull request includes a small refactor to the `GetExtra` method in `authn/auth_info.go`. 
The change populates the service identity in the extra map, even if an id token is involved in the query.
This allows services to check which service is making a call when receiving on-behalf-of user requests.


cinaglia (approved):
LGTM

PR #181, opened 2025-06-13
Implement retry when calling the TokenExchangeURL

filewalkwithme:
This PR introduces a retry logic when calling the external Token Exchange URL.

As of today, if the call fails because of a networking issue, the entire Exchange step fails too.

The current PR adds backoff mechanism that will retry the call for 3 times, with a max backoff interval of 4 seconds.

Fixes: https://github.com/grafana/search-and-storage-team/issues/326


authn/token_exchange.go
```diff
@@ -150,8 +151,28 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+			MaxBackoff: 4 * time.Second,
+			MinBackoff: time.Second,
+			MaxRetries: 3,
+		})
+
+		var res *http.Response
+		for b.Ongoing() {
+			res, err = c.client.Do(c.withHeaders(req))
+			// Retry the request if there was a fundamental error, like resolving the host or network error.
+			if err != nil {
```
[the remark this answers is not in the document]
filewalkwithme: Sometimes when we are making requests against a Kubernetes Service, with multiple pods behind, and one of the pods is being terminated, then we get an error like `dial tcp x.x.x.x:xxxx: connect: connection refused`

A retry would be helpful in such scenario, as another request would be made and it would end up (possibly) being served by another healthy pod.

Related to https://github.com/grafana/search-and-storage-team/issues/326

authn/token_exchange.go
```diff
@@ -150,8 +151,28 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
 			return nil, fmt.Errorf("failed to build http request: %w", err)
 		}
 
-		res, err := c.client.Do(c.withHeaders(req))
+		b := backoff.New(ctx, backoff.Config{
+			MaxBackoff: 4 * time.Second,
+			MinBackoff: time.Second,
+			MaxRetries: 3,
+		})
```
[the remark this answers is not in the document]
JohnnyQQQQ: The Loki client example you linked i.e. does that "correctly" by reusing the config over and over again.

authn/token_exchange.go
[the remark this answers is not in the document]
mihai-turdean: I agree, and that is a transient error. But I imagine there are other errors that are permanent, that should not be retried.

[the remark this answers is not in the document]
filewalkwithme: You are right! I pushed [a new commit](https://github.com/grafana/authlib/pull/181/commits/a82de401e431756c561948de1c2fdd5f64ef5082) modifying the retry logic to only retry on the following cases:
- 429 (rate limit exceeded)
- 500s (there's an issue with the server)
- Connection error

authn/token_exchange.go
```diff
@@ -53,6 +54,11 @@ func NewTokenExchangeClient(cfg TokenExchangeConfig, opts ...ExchangeClientOpts)
 		cache:   nil, // See below.
 		cfg:     cfg,
 		singlef: singleflight.Group{},
+		backoffCfg: backoff.Config{
+			MaxBackoff: 4 * time.Second,
+			MinBackoff: time.Second,
```
pstibrany, line 59: I would suggest to adjust min backoff to smaller time (250ms) in attempt to keep the latency down, and max backoff to 1s.

authn/token_exchange.go
```diff
@@ -145,13 +152,31 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
 
-		req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.cfg.TokenExchangeURL, bytes.NewReader(data))
-		if err != nil {
-			return nil, fmt.Errorf("failed to build http request: %w", err)
+		b := backoff.New(ctx, c.backoffCfg)
+
+		var req *http.Request
+		var res *http.Response
+		for b.Ongoing() {
+			req, err = http.NewRequestWithContext(ctx, http.MethodPost, c.cfg.TokenExchangeURL, bytes.NewReader(data))
```
pstibrany, line 160: Can we build the request once and reuse it for all the retries?

authn/token_exchange.go
```diff
@@ -145,13 +152,31 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to build http request: %w", err)
+			}
+
+			res, err = c.client.Do(c.withHeaders(req))
+			// Retry the request if there was a fundamental error, like resolving the host or network error,
+			// or if we get a 429 or a 500s HTTP status code
+			if err != nil || (res != nil && (res.StatusCode == http.StatusTooManyRequests || res.StatusCode >= http.StatusInternalServerError)) {
+				b.Wait()
```
pstibrany, line 169: Can we log the errors when we do the retry?

authn/token_exchange.go
```diff
@@ -145,13 +152,31 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+				continue
+			}
+
+			// No error, exit the retry loop
+			break
 		}
 
-		res, err := c.client.Do(c.withHeaders(req))
 		if err != nil {
+			// If we get here, it means we had hit the MaxRetries limit.
```
pstibrany, line 178: This is not necessarily true. Context might have been finished too. We should check for `b.Err()` here as well. (It could happen that context was canceled before first attempt, in which case `err` will be `nil`, but `b.Err()` will return error).

authn/token_exchange.go
```diff
@@ -145,13 +152,31 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+			// Only returns the last error.
 			return nil, fmt.Errorf("%w: %w", ErrInvalidExchangeResponse, err)
 		}
 		defer res.Body.Close()
```
pstibrany, line 182: We need to close response body after each attempt that ended with HTTP status codes that we retry.

authn/token_exchange.go
filewalkwithme: That's a very good point! Pushed a [new commit](https://github.com/grafana/authlib/pull/181/commits/a8d6b69c6e3356f1225171c537c4d416ad4c1118) to address that

authn/token_exchange.go
filewalkwithme: Thanks for catching that! Pushed a [new commit](https://github.com/grafana/authlib/pull/181/commits/282bf3e3480fcfa8a5fb69e21b922ac37dedc0be) handling this

gamab:
Thanks a lot 🙂 
Great work 💯

authn/token_exchange.go
```diff
@@ -163,16 +170,48 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+			res, err = c.client.Do(c.withHeaders(req))
+			// Retry the request if there was a fundamental error, like resolving the host or network error,
+			// or if we get a 429 or a 500s HTTP status code
+			if err != nil || (res != nil && (res.StatusCode == http.StatusTooManyRequests || res.StatusCode >= http.StatusInternalServerError)) {
+				b.Wait()
+
+				// Close response body after each attempt
+				if res != nil && res.Body != nil {
+					res.Body.Close()
+				}
```
pstibrany, line 192: Let's close this before calling `b.Wait()`, so that open connection can be put back to client's pool and reused by other goroutines.

authn/token_exchange.go
```diff
@@ -163,16 +170,48 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+		for b.Ongoing() {
+			req, err = http.NewRequestWithContext(ctx, http.MethodPost, c.cfg.TokenExchangeURL, bytes.NewReader(data))
+			if err != nil {
+				return nil, fmt.Errorf("failed to build http request: %w", err)
+			}
+
+			res, err = c.client.Do(c.withHeaders(req))
+			// Retry the request if there was a fundamental error, like resolving the host or network error,
+			// or if we get a 429 or a 500s HTTP status code
+			if err != nil || (res != nil && (res.StatusCode == http.StatusTooManyRequests || res.StatusCode >= http.StatusInternalServerError)) {
```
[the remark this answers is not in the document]
pstibrany: We have already checked `err != nil` in previous `if` condition. We don't need to check it again, and `res` must also be non-nil at this point.

authn/token_exchange.go
pstibrany: Good idea with the traces -- we could add event to the current span with error message. (I'm fine if we do this in separate PR)

authn/token_exchange.go
```diff
@@ -163,16 +170,48 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+				b.Wait()
+
+				// Close response body after each attempt
+				if res != nil && res.Body != nil {
```
pstibrany, line 190: `res.Body` is never `nil` when using http client.

authn/token_exchange.go
pstibrany: (It would be nice to test that we don't use more connections to server than necessary. We could use [counting dialer](https://github.com/grafana/backend-enterprise/blob/c5c281421f53a3c92cf49a7852e8c1463a282bc6/pkg/cloud/common/counting_dialer.go#L11), set it in http client, and then verify number of created connections)

pstibrany (approved):
Thank you. I'm fine merging this as-is, we can address remaining comments (if-checks, client conn reuse + test) in subsequent PR.

authn/token_exchange.go
filewalkwithme: Added a [new commit removing the unnecessary check](https://github.com/grafana/authlib/pull/181/commits/cb4cd01a203a563356b8cc5595b1d3e1c803272c)

authn/token_exchange.go
filewalkwithme: > Let's close this before calling b.Wait(), so that open connection can be put back to client's pool and reused by other goroutines. (In order for connection to be reused, we should also consume the body, eg. by using _, _ := io.Copy(io.Discard, res.Body))

Pushed a [new commit](https://github.com/grafana/authlib/pull/181/commits/57d14eb9cb87d252cab0ba51bc942d2719d92a4a) for that

authn/token_exchange.go
filewalkwithme: @gamab @pstibrany added a [new commit](https://github.com/grafana/authlib/pull/181/commits/d788dad84a1c25e5d9da819c462edf07111e918a) adding retry information to the Exchange span

authn/token_exchange.go
```diff
@@ -163,16 +171,51 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+			// while retrying the request (for example, context canceled).
+
+			// Returns Only the last error.
+			errRetry := err
+			if err == nil {
+				// in the absence of error, check error returned by the backoff struct
+				errRetry = b.Err()
+			}
+
+			return nil, fmt.Errorf("%w: %w", ErrInvalidExchangeResponse, errRetry)
```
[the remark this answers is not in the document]
JohnnyQQQQ: Changed

authn/token_exchange.go
```diff
@@ -204,6 +247,32 @@ func (c *TokenExchangeClient) Exchange(ctx context.Context, r TokenExchangeReque
...
+func addRetryInformationToSpan(span trace.Span, res *http.Response, err error) {
+	spanAttr := []attribute.KeyValue{}
+	if err != nil {
+		spanAttr = append(spanAttr, attribute.String("error", err.Error()))
+	}
+
+	if res != nil {
+		spanAttr = append(spanAttr, attribute.String("status", res.Status))
+	}
+	span.AddEvent("Failed HTTP request will be retried", trace.WithAttributes(spanAttr...))
```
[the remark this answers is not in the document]
pstibrany: In the latest change, `span.RecordError(err)` is already adding new event, so calling `span.AddEvent` later is unnecessary. (That’s why I added comment on that line, but I should have been more clear about it)

PR #182, opened 2025-06-13
AuthN: Account for traces feedback

gamab:


grpcutils/grpc_authenticator.go
```diff
@@ -65,7 +65,10 @@ func NewAuthenticatorInterceptor(auth authn.Authenticator, tracer trace.Tracer)
 
 		// FIXME: Add attribute with service subject once https://github.com/grafana/authlib/issues/139 is closed.
 		span.SetAttributes(attribute.String("subject", info.GetUID()))
-		span.SetAttributes(attribute.Bool("service", types.IsIdentityType(info.GetIdentityType(), types.TypeAccessPolicy)))
```
gamab, line 68: In the end I don't see what that brings that we wouldn't see thanks to the subject already.

IevaVasiljeva (approved):
LGTM

PR #191, opened 2025-09-24
AuthZ: Extract service permission evaluation from the client code

gamab:
In this PR, I extract the service permission evaluation from the client code to make it reusable to API level Authorizers that only need to check the incoming access token permissions.

### DIFF
--- a/authz/service_permissions.go
+++ b/authz/service_permissions.go
@@ -1,6 +1,9 @@
 package authz
 
 import (
+	"slices"
+	"strings"
+
 	"github.com/grafana/authlib/types"
 )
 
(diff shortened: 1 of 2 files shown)


authz/service_permissions.go
```diff
@@ -0,0 +1,66 @@
...
+	if res.ServiceCall {
+		res.Permissions = authInfo.GetTokenPermissions()
+	} else {
+		res.Permissions = authInfo.GetTokenDelegatedPermissions()
+	}
+	res.Allowed = hasPermissionInToken(res.Permissions, group, resource, verb)
+	return res
+}
+
+func hasPermissionInToken(tokenPermissions []string, group, resource, verb string) bool {
```
gamab, line 34: Unchanged, just moved.

PR #260, opened 2026-04-07
AuthN: Add serviceIdentity to actor claims and expose innermost to GetExtra

macabu:
This is part of the work to preserve the `serviceIdentity` through chained internally auth-ed requests.

It adds the `ServiceIdentity` into the `ActorClaims`, which will be populated in this PR: https://github.com/grafana/auth/pull/1232

And it also adds a `GetInnermostServiceIdentity` method to easily extract the initial service identity in a request chain.

### DIFF
--- a/authn/auth_info.go
+++ b/authn/auth_info.go
@@ -1,6 +1,7 @@
 package authn
 
 import (
+	"slices"
 	"strings"
 
 	"github.com/go-jose/go-jose/v4/jwt"
@@ -104,6 +105,17 @@ func (a *AuthInfo) GetExtra() map[string][]string {
 
 	if a.at.Rest.ServiceIdentity != "" {
 		result[ServiceIdentityKey] = []string{a.at.Rest.ServiceIdentity}
+	}
+
+	// This sets the original service identity used in an authenticated request chain.
+	// It will first look for the service identity in the actor claims, and if there's no actor,
+	// it uses the service identity from the subject's claim metadata.
+	// If the actor claims is present but the service identity was not passed or empty, that value is still the one returned.
+	// This is so we preserve the intent of the method, and always return the original identity even when empty.
+	if actor := a.at.Rest.getInnermostActor(); actor != nil {
+		result[InnermostServiceIdentityKey] = []string{actor.ServiceIdentity}
+	} else {
+		result[InnermostServiceIdentityKey] = slices.Clone(result[ServiceIdentityKey])
 	}
 
 	return result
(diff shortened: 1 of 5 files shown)


authn/auth_info_test.go
```diff
@@ -543,3 +543,75 @@ func TestAuthInfo_GetIdentifier(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			expected: "innermost-service-identity",
+		},
+		{
+			name: "with an access token with an actor, returns the service identity from the actor even if it is empty",
```
macabu: Is this a behavior we want? For the example request chain:
SvcA -> SvcB -> SvcC

If `SvcA` does not set a `ServiceIdentity`, and `SvcB` does, do we want to return empty here? I think that makes sense...

The point is that it is still about the original service identity, otherwise we would have inconsistent behavior, where sometimes it would be the first service in the chain, and other times not depending on whether they have `ServiceIdentity` set.

authn/auth_info.go
```diff
@@ -180,3 +180,11 @@ func (a *AuthInfo) GetIDToken() string {
 func (a *AuthInfo) GetAccessToken() string {
 	return ""
 }
+
+func (a *AuthInfo) GetInnermostServiceIdentity() string {
```
macabu: Should I add this method to the `types/auth.go` interface? Also would welcome suggestions about the name, but would like to differentiate the intent here that we want the **innermost** service identity, not just the service identity in general.

macabu (commented):
Self-review with questions!

macabu:
@codex review

authn/auth_info.go
```diff
@@ -180,3 +180,11 @@ func (a *AuthInfo) GetIDToken() string {
...
+	if actor := a.at.Rest.getInnermostActor(); actor != nil {
+		return actor.ServiceIdentity
```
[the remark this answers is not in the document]
macabu: Yes this is one of the questions we're wondering about the implementation. I think we can document the behavior but accept it, it feels better than inconsistent behavior.

authn/auth_info.go
macabu: Added the method to the interface, otherwise it becomes a bit weird to use having to cast to a concrete struct.

macabu: Removing in favor of using `GetExtra()`

cinaglia (approved):
LGTM

PR #261, opened 2026-04-09
authn: key token exchange cache by stable subject + actor chain

charandas:
## Summary
- change token exchange request hashing to derive cache key from a stable subject identity rather than raw subject token bytes
- reuse existing recursive ActorClaims type and include flattened actor lineage (top-level sub + nested act.sub chain) in the cache key
- preserve backward-compatible fallback to raw subjectToken when parsing or claims extraction fails, or stable fields are absent
- add tests to verify cache reuse across token refresh with same actor chain and cache misses when nested actor changes

## Verification
- go test ./authn -run TokenExchange -count=1

### DIFF
--- a/authn/token_exchange.go
+++ b/authn/token_exchange.go
@@ -179,6 +179,10 @@ func flattenedActorSubjects(actor *ActorClaims) []string {
 		}
 		actor = actor.Actor
 	}
+	// Canonicalize to originator-first order (deepest actor first).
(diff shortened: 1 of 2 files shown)


mihai-turdean (approved):

mgyongyosi (approved):

cinaglia (approved):
LGTM

PR #264, opened 2026-04-30
AuthN: Add optional Groups to IDTokenClaims

mgyongyosi:
## Summary
- Adds an optional `Groups []string` field to `IDTokenClaims` in `authn/verifier_id_token.go`, populated from the identity's group attribute.
- Wires `AuthInfo.GetGroups()` to return the ID token groups, falling back to the innermost actor's groups when no ID token is present.
- Adds `TestAuthInfo_GetGroups` covering ID-token, actor, and nested-actor paths.

Relates to https://github.com/grafana/grafana/pull/123249

## Test plan
- [ ] `go test ./authn/...`
- [ ] Verify ID tokens issued with group memberships expose `groups` in the parsed claims and that `AuthInfo.GetGroups()` returns them
- [ ] Verify access tokens with an inner actor (no ID token) surface the innermost actor's groups via `GetGroups()`

### DIFF
--- a/authn/auth_info.go
+++ b/authn/auth_info.go
@@ -91,7 +91,10 @@ func (a *AuthInfo) GetNamespace() string {
 }
 
 func (a *AuthInfo) GetGroups() []string {
-	return []string{}
+	if a.id != nil {
+		return a.id.Rest.Groups
+	}
+	return nil
 }
 
 func (a *AuthInfo) GetExtra() map[string][]string {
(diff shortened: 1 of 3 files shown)


chaudyg (approved):

authn/auth_info.go
```diff
@@ -91,7 +91,10 @@ func (a *AuthInfo) GetNamespace() string {
 }
 
 func (a *AuthInfo) GetGroups() []string {
-	return []string{}
+	if a.id != nil {
```
chaudyg, line 94: could you add a test ?

mgyongyosi: Done, also changed to support OBO access tokens.

PR #286, opened 2026-06-01
authn: Add request_headers and response_body response fields

jtroy:
This PR adds two new fields to the authn-server response:
- `request_headers` are the headers which the HG gateway will tack onto the request to be sent downstream to the destination multi-tenant service when authentication succeeds. This will include `Authorization`, `X-Access-Token`, etc.
- `response_body` is the response body which HG will send to the client when authentication fails. This is needed today to communicate token revocation to the frontend.

`request_headers` subsumes the existing token-returning functionality, so `token` has been marked deprecated and will be removed when the HG gateway and authn service are updated.


cinaglia (approved):
LGTM

Rydez (approved):

PR #296, opened 2026-06-05
feat: add id token signer client

cbodonnell:
This PR adds an `IDTokenSigner` client to `authlib` that calls the auth API's `/v1/sign-id-token` endpoint. It enables services to mint ID tokens for on-behalf-of authentication flows without having to implement the HTTP client logic themselves.

PR #298, opened 2026-06-10
feat(authn): add Extra claims and configurable headers to IDTokenSigner

cinaglia:
Extends the ID token signer client so it can be reused by services that need to pass identity claims (email, username, type, etc.) and customize the request headers for stack-scoped callers.

> **Note:** This PR is meant to be merged into #296 first.

**Changes:**
- Add `Extra map[string]any` to `SignIDTokenRequest` for additional token claims
- Add `WithIDTokenSignerHeaders` option to override default system-scoped headers
- Cache key accounts for `Extra` to avoid collisions

Made with [Cursor](https://cursor.com)


cinaglia:
Closing as the commit has been cherry-picked in https://github.com/grafana/authlib/pull/296.

PR #299, opened 2026-06-14
feat(authn): Support subject identity in token exchange request

cinaglia:
## Why

Today, obtaining an on-behalf-of (OBO) access token through the exchange client requires a `subjectToken`, either an existing access token or an ID token. Callers that only hold a user identity must first sign an ID token and then exchange it, which is two round-trips and encourages standalone ID-token minting.

This lets callers pass a `subject` identity directly on the exchange request, so an identity-bearing access token can be obtained in one step, with no intermediate ID token required.

## What

- Add an optional `Subject` field to `TokenExchangeRequest`, backed by a new `TokenExchangeSubject` type whose JSON shape mirrors the sign-access-token `subject` object (`identifier`, `type`, `namespace`, `authenticatedBy`, `email`, `email_verified`, `username`, `name`, `role`, `groups`).
- `Subject` and `SubjectToken` are mutually exclusive; setting both returns the new `ErrMutuallyExclusiveSubject` error.
- Fold the subject into the request cache key so tokens for different identities are not shared.
- Add tests covering request propagation, mutual exclusivity, and per-subject cache isolation.

### DIFF
--- a/authn/token_exchange.go
+++ b/authn/token_exchange.go
@@ -117,12 +117,34 @@ type TokenExchangeRequest struct {
 	Audiences []string `json:"audiences"`
 	// [Optional] SubjectToken is the token to exchange in case of a token exchange request.
 	SubjectToken string `json:"subjectToken,omitempty"`
+	// [Optional] Subject is a user identity to sign an on-behalf-of token for,
+	// without first obtaining a signed subjectToken. It is mutually exclusive
+	// with SubjectToken. The auth API requires the caller to hold the
+	// grafana-id-token:sign scope (in addition to access-token:sign) to use it.
+	Subject *TokenExchangeSubject `json:"subject,omitempty"`
 	// [Optional] ExpiresIn is the duration, in seconds, before the token expires.
 	ExpiresIn *int `json:"expiresIn,omitempty"`
 	// [Optional] RestrictedDelegatedPermissions narrows the token's delegated permissions
 	// to the intersection of this list and the CAP's delegated permissions. If omitted,
 	// the full set of CAP delegated permissions is used.
 	RestrictedDelegatedPermissions []string `json:"restrictedDelegatedPermissions,omitempty"`
+}
+
+// TokenExchangeSubject carries a user identity that the caller asserts when
+// exchanging for an on-behalf-of access token. Its JSON shape matches the
+// auth API's sign-access-token "subject" object and mirrors the claim set of
+// an ID token.
+type TokenExchangeSubject struct {
+	Identifier      string   `json:"identifier"`
+	Type            string   `json:"type"`
+	Namespace       string   `json:"namespace"`
+	AuthenticatedBy string   `json:"authenticatedBy,omitempty"`
+	Email           string   `json:"email,omitempty"`
+	EmailVerified   bool     `json:"email_verified,omitempty"`
+	Username        string   `json:"username,omitempty"`
+	DisplayName     string   `json:"name,omitempty"`
+	Role            string   `json:"role,omitempty"`
+	Groups          []string `json:"groups,omitempty"`
 }
 
 type TokenExchangeResponse struct {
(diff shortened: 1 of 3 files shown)


volcanonoodle (dismissed):

charandas (dismissed):

authn/token_exchange.go
```diff
@@ -136,6 +158,14 @@ func (r TokenExchangeRequest) hash() string {
 	sort.Strings(r.Audiences)
 	br.WriteString(strings.Join(r.Audiences, "-"))
 	br.WriteString(subjectCacheKey(r.SubjectToken))
+	if r.Subject != nil {
+		// json.Marshal of a struct is deterministic (field order is stable), so
+		// this yields a stable cache key that differs per subject identity.
+		if data, err := json.Marshal(r.Subject); err == nil {
+			br.WriteByte('-')
+			br.Write(data)
```
Rydez: will the groups always be in the same order as well?

Rydez (commented):
looking good. just one question.

authn/token_exchange.go
```diff
@@ -125,6 +130,23 @@ type TokenExchangeRequest struct {
...
+	Identifier      string   `json:"identifier"`
+	Type            string   `json:"type"`
+	Namespace       string   `json:"namespace"`
+	AuthenticatedBy string   `json:"authenticatedBy,omitempty"`
+	Email           string   `json:"email,omitempty"`
+	EmailVerified   bool     `json:"email_verified,omitempty"`
+	Username        string   `json:"username,omitempty"`
+	DisplayName     string   `json:"name,omitempty"`
+	Role            string   `json:"role,omitempty"`
+	Groups          []string `json:"groups,omitempty"`
```
jtroy, line 147: Should `Groups` be sorted before we `hash()`?

authn/token_exchange.go
```diff
@@ -136,6 +158,14 @@ func (r TokenExchangeRequest) hash() string {
 	sort.Strings(r.Audiences)
 	br.WriteString(strings.Join(r.Audiences, "-"))
 	br.WriteString(subjectCacheKey(r.SubjectToken))
+	if r.Subject != nil {
+		// json.Marshal of a struct is deterministic (field order is stable), so
+		// this yields a stable cache key that differs per subject identity.
+		if data, err := json.Marshal(r.Subject); err == nil {
```
jtroy: nit: it looks like this `Marshal()` can't fail today, but maybe it's worth handling this error defensively in case we ever add something to the subject that can fail to marshal. The potential bug is that it's impossible to tell the difference between a request with no `Subject` and an otherwise-identical one with a `Subject` that failed to marshal.

authn/token_exchange.go
jtroy: nvm, @Rydez already mentioned this

authn/token_exchange.go
cinaglia: Good catch. Addressed via [a8be3c1](https://github.com/grafana/authlib/pull/299/commits/a8be3c1bb96f264844d94ff7988a10735586e8f8). PTAL.

authn/token_exchange.go
cinaglia: Good feedback. Addressed via [a3c4eab](https://github.com/grafana/authlib/pull/299/commits/a3c4eab7a4eb06599a25ba894bb153e5185d8121). PTAL.

authn/token_exchange.go
jtroy: I realize you didn't touch this, but if we're going to do the `copy()` ceremony on the other slices to avoid modifying them under the caller, I think we should do that here too.

jtroy (dismissed):
Looks good! One nit above about making existing code match, but approve and leave that up to your judgment.

authn/token_exchange.go
cinaglia: Good call. Addressed via [1452740](https://github.com/grafana/authlib/pull/299/commits/1452740b7a0bfae15624cd6f32da3c5ed3509ecb).

Rydez (approved):

jtroy (approved):

PR #326, opened 2026-09-19
id-token-deriver: new interface and implementation

charandas:
Builds a new interface and implementation on top of upcoming API in our auth deployment. The cache, retry and backoff mechanism was pulled out of TokenExchangeClient, so it could be reused by the IDTokenDeriveClient.


authn/id_token_deriver.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/grafana/authlib/internal/httpclient"
+	"github.com/grafana/dskit/backoff"
+)
+
+// IDTokenDeriver mints an id token identifying the user or service account embedded in an
+// on-behalf-of access token's actor chain. It is for callers that hold an access token instead of
+// a directly-asserted subject; compare IDTokenSigner, which signs a directly-asserted subject.
+type IDTokenDeriver interface {
+	DeriveIDToken(ctx context.Context, subjectToken, namespace string) (*DeriveIDTokenResponse, error)
+}
```
ryantxu, line 25:
```suggestion
type IDTokenMinter interface {
	MintIDToken(ctx context.Context, subjectToken, namespace string) (*DeriveIDTokenResponse, error)
}
```

maybe 🤷🏻  -- IIUC, this is what we already call it colloquially, it took me a while to figure out what the role of a IDTokenDeriver is 😛

charandas: Already discussed in slack, that derive is the term that was chosen with the identity team.

authn/token_cache.go
```diff
@@ -0,0 +1,189 @@
...
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/sync/singleflight"
+
+	"github.com/grafana/authlib/cache"
+	"github.com/grafana/dskit/backoff"
+)
+
+// authAPIEnvelope is the {data:{token},status,error} response shape shared by every auth-api
+// token-minting endpoint (sign-access-token, exchange, derive-id-token).
+type authAPIEnvelope struct {
```
cinaglia: nit/optional:

```suggestion
type authAPIResponse struct {
```

authn/id_token_deriver.go
```diff
@@ -0,0 +1,157 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+
+	"github.com/grafana/authlib/cache"
+	"github.com/grafana/authlib/internal/httpclient"
+	"github.com/grafana/dskit/backoff"
+)
+
+// IDTokenDeriver mints an id token identifying the user or service account embedded in an
```
cinaglia: Perhaps not using the word "mint" here would help avoid some confusion. Perhaps we can say that it is responsible for extracting an identity from an on-behalf-of access token actor chain and returning a signed ID token.

authn/token_cache.go
cinaglia: nit:
```suggestion
type authAPIResponse struct {
```

authn/token_cache.go
```diff
@@ -0,0 +1,189 @@
...
+		return "", fmt.Errorf("%w: %w", errInvalid, errors.Join(b.Err(), err))
+	}
+
+	if res.StatusCode >= http.StatusInternalServerError {
+		return "", fmt.Errorf("%w: %s", errInvalid, res.Status)
+	}
+
+	response := authAPIEnvelope{}
+	if err := json.NewDecoder(res.Body).Decode(&response); err != nil {
+		return "", err
```
cinaglia: Probably not worth tackling this now, but I wonder why we're not wrapping this error in `errInvalid`.

authn/token_cache.go
```diff
@@ -0,0 +1,189 @@
...
+	Status string `json:"status"`
+	Error  string `json:"error"`
+}
+
+// tokenCache mints and caches short-lived bearer tokens from auth-api, keyed by an opaque cache
+// key the caller derives from its own request shape. Concurrent calls for the same key are
+// coalesced via singleflight, and cache TTL is taken from each minted token's own JWT expiry.
+// Shared by every auth-api token client in this package so the retry/backoff/cache plumbing is
+// written once instead of once per capability (exchange, derive, ...).
+type tokenCache struct {
```
cinaglia: I like the idea of sharing as much as possible between `TokenExchangeClient` and `IDTokenDeriveClient`. I wonder if placing most of this functionality under "token_cache" is the way to go, though. It seems like we could split this work into two parts: one for caching and another one for the HTTP API client.

Could we perhaps update `token_cache.go` to only keep caching: `tokenCache` plus its lookup method. Perhaps rename it to `tokenFor(ctx, key, request)` (or keep `getOrRequest`). 

Then, we could introduce `auth_api_client.go`, which would hold the shared Auth API HTTP related things (request/retry/backoff/decode/etc).

cinaglia (dismissed):
I think this is looking pretty good. I've left a few comments but nothing blocking.

authn/token_cache.go
charandas: Implemented in https://github.com/grafana/authlib/pull/326/commits/69d965830d1403ed0cc38e2db8c353bcebd09376

authn/token_cache.go
charandas: Addressed in https://github.com/grafana/authlib/pull/326/commits/69d965830d1403ed0cc38e2db8c353bcebd09376

authn/token_cache.go
charandas: Done in https://github.com/grafana/authlib/pull/326/commits/69d965830d1403ed0cc38e2db8c353bcebd09376

authn/id_token_deriver.go
charandas: Addressed: https://github.com/grafana/authlib/pull/326/commits/69d965830d1403ed0cc38e2db8c353bcebd09376

cinaglia (approved):
LGTM

PR #327, opened 2026-09-20
fix(authn): avoid caching tokens within the expiry margin

kolesnikovae:
Tokens received with 15 seconds or less remaining can be cached indefinitely: subtracting the existing refresh margin produces a non-positive TTL, which the default cache treats as no expiration. Skip caching those tokens.

Discovered while investigating an unrelated issue. This is unlikely in normal use; default token lifetimes are safely above the margin.

Validated with deterministic boundary tests and race-enabled authn/authz tests.


kolesnikovae:
cursor review

forsethc (approved):

cinaglia (approved):
LGTM. Thanks!

